PR #1, opened 2022-08-02
add almanac updater strawman

evanmoss:
- @punov can you set it up so the fixture directory gets copied into the compiled code dir so it's available at runtime in the directory the server is run from?
- I haven't tested the output part of ubxtool since I'm doing this blind, I have no idea what the output will look like, but assuming comma-delimited bytes like that 🤷


punov (approved):
LGTM, thank you!
fixtures/ folder is now copied over to compiled directory,
we'll test `almanac.ts` soon in action

PR #6, opened 2022-10-25
Ingest GNSS Jamming & Spoofing States

BenDavidAaron:


BenDavidAaron:
The `/api/1/spoofdetstate` and `/api/1/jamind` routes are available and ready to ingest. 

I've only been able to query `NaN` from the route for the value of `spoofDetState`, I don't have any reason to believe it wont work since I can get a value out of `ubxtool`. 
We should test that with an HDC outside, with a GPS fix, somewhere that isn't Ben's Desk, before fully accepting this PR. I can load up the camera API and possibly test it tomorrow with John?

BenDavidAaron:
I've been working on configuring the GNSS chip to show a value for `spoofDetState`, so far `ubxtool -z CFG-ITFM-ENABLE 1` isn't the right invocation to configure it, but I'm betting it's described in the interface doc and I'm gonna figure it out.

BenDavidAaron:
`ubxtool -z CFG-ITFM-ENABLE,1` is the invocation to enable IFTM

BenDavidAaron:
Per a conversation I just had with Nathaniel:

This GNSS chip configuration has been pushed as far as I think I'm capable of pushing it. We need Hellbender to step in and do some EE magic on setting the `CFG-ITFM-ENABLE` to `1`. 

The API route currently returns `null` (since the `spoofDetState` has not been enable-able). It _should_ return an integer value (0, 1, 2) based on the spoof state. I'm using a pattern from the `api/1/jamInd` route, so there's no reason it shouldn't work once `spoofDetState` is properly configured (assuming that the `ubxtool` output follows the same format).

punov (approved):

AriBruck:
@BenDavidAaron a note regarding possible values
`spoofDetState` has 4 possible values:
- -
Spoofing detection state (not supported for protocol U:2 spoofDetState
versions less than 18.00)
• 0:Unknownordeactivated
• 1:Nospoofingindicated
• 2:Spoofingindicated
• 3:Multiplespoofingindications

Note that the spoofing state value only reflects the detector state for the current navigation epoch. As spoofing can be detected most easily at the transition from real signal to spoofing signal, this is also where the detector is triggered the most. I.e. a value of 1 - No spoofing indicated does not mean that the receiver is not spoofed, it simply states that the detector was not triggered in this epoch.

punov:
Closed as outdated

PR #15, opened 2023-01-10
Adding Publish a package we can ingest into the firmware

hmn8te:
WIP; don't merge yet please


hmn8te:
This is a WIP as I try to make github actions work

PR #26, opened 2023-01-19
Expose Resolution endpoint for camera config

h-tohkubo:
Enable camera resolution configuration for 4K, 2K, 1K options.
* expose an endpoint to configure properties for w and h based on resolution type
* integrated jest with supertest to test express api endpoints
* ran `prettier` for formatting


tsconfig.json
```diff
@@ -45,7 +45,12 @@
     /* Module Resolution Options */
     // "moduleResolution": "node",                  /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
     "baseUrl": "./src" /* Base directory to resolve non-absolute module names. */,
-    // "paths": {},                                 /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */
+    "paths": {
+      "@config/*": ["config/*"],
+      "@routes/*": ["routes/*"],
+      "@services/*": ["services/*"],
+      "@util/*": ["util/*"]
+    } /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */,
```
h-tohkubo, line 53: I may remove this since `baseUrl` is already `./src`

src/index.ts
```diff
@@ -1,82 +1,49 @@
...
+  busboy({
+    highWaterMark: 2 * 1024 * 1024, // Set 2MiB buffer
+  }),
+); // Handles file uploads for Over-The-Air update
+app.use(express.json());
+app.use(express.urlencoded({ extended: true }));
+app.use(router);
 
+(async () => {
   await new Promise<void>((resolve, reject) => {
```
siiiiilvio, line 26: @punov This line looks suspicious. Is there a need to wrap this synchronous operation with a Promise that will resolve right away? 

I think the only effect would be to send this operation to the event loop (micro). In other words, wait until current stack execution is finished, then execute this line. Should it just go at the end then?

src/services/gnssHealthCheck.ts
```diff
@@ -15,7 +15,7 @@ export const GnssHealthCheck: IService = {
 
           if (cmdOutput.indexOf('active') !== 0) {
             exec('systemctl start gpsd');
-            // console.log('GPSD was down. Bringing back!');
+            console.log('GPSD was down. Bringing back!');
```
hmn8te, line 18: This is an unrelated change. This should go in a separate PR.

src/services/gnssHealthCheck.ts
```diff
@@ -29,7 +29,7 @@ export const GnssHealthCheck: IService = {
 
               if (cmdOutput.indexOf('active') !== 0) {
                 exec('systemctl start gnss-logger');
-                // console.log('Gnss-logger was down. Bringing back!');
+                console.log('Gnss-logger was down. Bringing back!');
```
hmn8te, line 32: Same as above.

hmn8te (changes requested):
I'm going to be annoying, because this is a larger PR and it could be much smaller and faster to merge. 

Formatting should be a separate PR. There are 17 commits and 17 files changes. That's a fair amount for adding an endpoint that accepts a single parameter.

I think we can rebase this and split things up pretty quickly?

src/index.ts
hmn8te: Is this change related to the new endpoint at all @h-tohkubo?

h-tohkubo:
Decided against decoupling resolution logic as a separate endpoint. Closing PR.

PR #28, opened 2023-01-20
Remove check for time set on restart camera

h-tohkubo:


siiiiilvio (approved):

src/services/updateCameraConfig.ts
```diff
@@ -19,22 +18,19 @@ export const UpdateCameraConfigService: IService = {
...
     } catch (e: unknown) {
       console.log('Camera Config service failed with error', e);
     }
   },
   delay: 200,
 };
+
+const restartCamera = () => {
+  exec(CMD.STOP_CAMERA, () => {
+    exec(CMD.START_CAMERA, () => {
```
punov, line 32: @h-tohkubo 
can we add 1-2 sec sleep before executing start?
I noticed how scheduling commands like this one RIGHT after another sometimes brings issues on Pi
like, system isn't fully got that process stopped, and we're kicking start

punov (changes requested):

src/services/updateCameraConfig.ts
h-tohkubo: Yup! Added sleep.

punov (approved):

PR #37, opened 2023-01-27
GET current configurations from local

h-tohkubo:
* Expose new endpoint for GETting existing configurations from odc-api's local store
* Add jest with supertest to verify both GET and POST `/config/cameraconfig`


siiiiilvio (approved):
👍

PR #43, opened 2023-02-06
Stefan/preview

ste00martin:
adding the /tmp/public directory to be available for any public static files; this has been tested and verified as functional on a dashcam

http://192.168.0.10:5000/tmp/1675454627.jpg - a new public endpoint of preview images


ste00martin:
Do not merge - this may be functional but we want to test further before merging into main

hmn8te:
@ste00martin One thought I had is that removing the ODC API breaks any build outside of github actions. I'm okay with that, but wanted to call it out.

siiiiilvio (approved):
👍

punov (approved):

PR #67, opened 2023-05-18
Image preview: Sort images based on timestamp and return the two newest.

hmn8te:
Updating the preview logic to show the most recent two images. 

The HDC-S exposed an issue that the previous preview logic was relying on the fact that the files names where in order. With the HDC-S this was not true. Shift the logic to use the timestamp of the file.


hmn8te:
Please don't merge until I test.

punov (approved):

PR #85, opened 2023-08-28
API to swap model

punov:


src/config/hdc-s.ts
```diff
@@ -45,6 +45,9 @@ export const CACHED_RES_CONFIG = '/data/res.config';
 export const UPLOAD_PATH = __dirname + '/../../../data/';
 export const DEVICE_INFO_LOG_FILE = __dirname + '/../../../tmp/dump.bin';
 export const CRON_CONFIG = '/home/root/cron_config';
+export const ML_MODELS = {
+  PVC: '/data/models/pvc.onnx'
```
ste00martin, line 49: why is it not the same path as hdc?

ste00martin (approved):
only ask is to fix the package.json dependencies

ste00martin (changes requested):
can you push your package-lock.json changes also?

src/routes/ml.ts
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * To swap model,
+   * 1. Get current model hash: GET /ml/hash?type=PVC
+   * 2. Upload new model to /tmp/pvc.onnx using POST /upload
+   * 3. Commit new model: POST /ml/commit with body { type: 'PVC', hash: 'new hash', path: '/tmp/pvc.onnx' }
+   */
+  router.get('/hash', async (req: Request, res: Response) => {
+    try {
+      if (!req.query.type) {
+        res.json({ error: 'model type is required' });
```
ste00martin, line 18: what does this mean? model type?

src/routes/ml.ts
```diff
@@ -0,0 +1,55 @@
...
+        return;
+      }
+      // @ts-ignore
+      const modelPath = ML_MODELS[req.query.type];
+      if (!modelPath) {
+        res.json({ error: 'model type is not supported' });
+        return;
+      }
+      const hash = readFileSync(modelPath + '.hash', { encoding: 'utf-8' });
+      res.json({ hash: hash.trim() } );
```
ste00martin, line 28: can we get an empty string if there's nothing we get back?
```suggestion
      const hash = readFileSync(modelPath + '.hash', { encoding: 'utf-8' });
      res.json({ hash: hash.trim() } );
```

ste00martin (approved):
can you also add package-lock.json?

PR #95, opened 2023-10-03
added service to write images from USB Stick to respective date direc…

HariHive:
…tory


src/services/saveImagesToUSB.ts
```diff
@@ -0,0 +1,54 @@
...
+import { DiskUsage } from 'types/motionModel';
+import { readdirSync, existsSync, mkdirSync, rename } from 'fs';
+import { USB_WRITE_PATH } from 'config';
+import { getDateFromUnicodeTimestamp } from '../util';
+
+const diskUsage: DiskUsage = {};
+
+export const getDiskUsage = () => {
+    return diskUsage;
+};
```
punov, line 11: @HariHive this is a leftover from another service, right?

src/services/saveImagesToUSB.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const SaveImagesToUSB: IService = {
+    execute: async () => {
+        try {
+            const directoryPath = USB_WRITE_PATH;
+            const files = readdirSync(directoryPath);
```
punov, line 17: I need to double-confirm that if this request is very long (for example, if there're lots of images in directory), it will not hang the entire ODC API.
And we can test it the following way:
- Let your camera to create a looooot of frames on your USB stick, just run the camera and let it create thousands of them, without moving to any folders
- Then, update dashcam-api.js on the camera with your new solution, and try to execute it, so it will try to move all those thousands of images to folders
- Check that API itself is alive and working during the execution of your heavy request (you can check if http://192.168.0.10:5000/api/1/info is still responsive)
- If API is alive, and your script is capable of moving thousands of images to sub-folders, then we're ready to go

src/services/saveImagesToUSB.ts
```diff
@@ -0,0 +1,54 @@
...
+
+            files.forEach((file) => {
+                if (file.endsWith("jpg")) {
+                    const dateFromFileName = getDateFromUnicodeTimestamp(file);
+                    const year = dateFromFileName.getFullYear();
+                    const month = String(dateFromFileName.getMonth() + 1).padStart(2, "0"); // Add 1 because months are 0-based
+                    const day = String(dateFromFileName.getDate()).padStart(2, "0");
+
+                    // Create the formatted date string
+                    const formattedDate = `${year}-${month}-${day}`;
```
punov, line 27: There's also an easy way to get this from the date object like:
`new Date().toISOString().split('T')[0]`
that will give you YYYY-MM-DD

src/services/saveImagesToUSB.ts
```diff
@@ -0,0 +1,54 @@
...
+                            mkdirSync(directoryPath, { recursive: true });
+                        } catch (error) {
+                            console.error(`FROM SAVEIMAGESTOUSB SERVICE :::: Error creating directory '${directoryPath}':`, error);
+                        }
+                    }
+
+                    //Move the JPG image file to the right directory
+                    const sourcePath = `${USB_WRITE_PATH}/${file}`;
+                    const destinationPath = `${USB_WRITE_PATH}/${formattedDate}/${file}`;
+                    rename(sourcePath, destinationPath, (err) => {
```
punov, line 42: When you will be executing this over thousands of images,
let's also measure, how much time the entire operation will take!
Just doing
Date.now() at the start, and comparing with Date.now() at the end will be good enough

punov (changes requested):

src/services/saveImagesToUSB.ts
HariHive: yes right, removed

src/services/saveImagesToUSB.ts
HariHive: updated!

src/services/saveImagesToUSB.ts
```diff
@@ -0,0 +1,49 @@
...
+            //Check if USB is connected
+            const usbConnected = existsSync(USB_WRITE_PATH);
+            if (usbConnected) {
+                const directoryPath = USB_WRITE_PATH;
+                const files = readdir(directoryPath, (err, files) => {
+                    files.forEach((file) => {
+                        if (file.endsWith("jpg")) {
+
+                            const formattedDate = new Date().toISOString().split('T')[0];
+
```
punov, line 17: Sorry for confusion, @HariHive, but I still mean you should take the date from the file name,
I just shared an easy way to get YYYY-MM-DD from it,
your approach with `getDateFromUnicodeTimestamp` was right.

src/services/saveImagesToUSB.ts
```diff
@@ -0,0 +1,49 @@
...
+                            const directoryPath = `${USB_WRITE_PATH}/${formattedDate}`;
+
+                            //Check if directory exists for the day on which image was collected
+                            if (!existsSync(directoryPath)) {
+                                try {
+                                    mkdirSync(directoryPath, { recursive: true });
```
punov, line 23: Don't think `{ recursive: true }` is needed here, but not critical

src/services/saveImagesToUSB.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const SaveImagesToUSB: IService = {
+    execute: async () => {
+        try {
+            //Check if USB is connected
+            const usbConnected = existsSync(USB_WRITE_PATH);
+            if (usbConnected) {
+                const directoryPath = USB_WRITE_PATH;
+                const files = readdir(directoryPath, (err, files) => {
+                    files.forEach((file) => {
```
punov, line 13: Correct me if I'm wrong, but forEach will execute all of them pretty much at the same time.
Not sure if it's healthy for CPU or not,

I'd do a regular for-loop and promisify `rename` non-blocking function,
(you can take a look on usage of `promisify` in project),
and then do 
for () {
await promisifiedRename...
}

this way for-loop + await should still be non-blocking, but you will also execute those operations one-by-one, not at the same time.

punov (approved):

src/services/saveImagesToUSB.ts
HariHive: removed!

src/services/saveImagesToUSB.ts
HariHive: updated

evanmoss:
@HariHive @punov why is this diff so large?   Please make sure you've read the new branching, testing, merging instructions 

Screen Shot 2023-10-03 at 6 25 24 PM

src/services/saveImagesToUSB.ts
```diff
@@ -0,0 +1,49 @@
+import { IService } from '../types';
+import { readdirSync, existsSync, mkdirSync, rename, statSync, readdir } from 'fs';
+import { USB_WRITE_PATH } from 'config';
+
+export const SaveImagesToUSB: IService = {
```
evanmoss, line 5: are we saving images to the usb, or are we moving images to a new directory?  It's really important that we are clear and accurate with our function names and descriptions.

src/services/saveImagesToUSB.ts
```diff
@@ -0,0 +1,49 @@
...
+    execute: async () => {
+        try {
+            //Check if USB is connected
+            const usbConnected = existsSync(USB_WRITE_PATH);
```
evanmoss, line 9: how often is this called?  should the existence of the USB be a given for this to be called in the first place?  I.e., shouldn't we move that upstream of this function?

src/services/saveImagesToUSB.ts
```diff
@@ -0,0 +1,49 @@
...
+                    if (err) {
+                        console.log("Failed to read directory frm serviceeeeeee");
+                    }
+                });
+            }
+        } catch (error) {
+            console.error('Error:', error);
+        }
+    },
+    interval: 118888,
```
evanmoss, line 48: @punov where does this come from?  What does it mean?

src/services/saveImagesToUSB.ts
evanmoss: This is is simply creating a new timestamp of today's date, but we need to accomplish moving the file to the day folder that matches with its timestamp.

evanmoss (changes requested):
@HariHive @punov A few callouts:
- The spec calls for moving files to a folder representing the day they were created (think of the user trying to find data from a certain day)
- We need to be accurate and precise in our naming of things
- We should prefer to use async/await instead of callbacks
- We should stick to established formatting norms (in this case the tab spacing, for example)
- Prefer using builtin path/os methods like `path.join()`
- I would prefer using utility functions for common general operations like getting day strings from dates, etc.  We can copy paste things that already exist in the network repo under server/src/util/index.ts

It's hard for me to see the whole diff or understand when this runs because of the large diff.

In terms of approach, I think this is a nice first attempt, but I think we can do better by pushing more to the OS.

What are we trying to accomplish here?  1. list files that haven't been put in correct directoris yet, 2: put them in the correct directories.

Let's redesign our code to have 2 helper functions:

1. a function that returns the list of files that are not in the proper directories
2. If we're worried about OS I/O, a function that can move files in bulk would be preferable to reduce the OS I/O from O(n) to O(1) -- I think instead of a library, we can use a simple child_process.exec('mv <something>')

how are the files names today?  If they are by timestamp then we can use a wildcard, otherwise, we can try `mv ${filesToMove.join(' ')} ${destination}`.  If we are worried about the command line argument size growing too large, we can batch this up into chunks.

src/services/saveImagesToUSB.ts
punov: Agree on this one. 
Though contributor can technically disconnect USB device on-the-go (rare case)
Do you think it's expensive to double-check the existance of this folder before running?

src/services/saveImagesToUSB.ts
punov: Feels like Hari didn't commit the fix for this line, 
she initially was using the date from the file name

punov:
> @HariHive @punov why is this diff so large? Please make sure you've read the new branching, testing, merging instructions
> 
> Screen Shot 2023-10-03 at 6 25 24 PM

@HariHive can you please add a compiled file to .gitignore?

src/services/saveImagesToUSB.ts
punov: @HariHive still there, maybe you didn't actually commit the latest?

punov (changes requested):

src/services/moveImagesOnUSB.ts
```diff
@@ -0,0 +1,77 @@
...
+                        mkdirSync(path.join(USB_WRITE_PATH, formattedDate));
+                        DIRS_EXISTING.add(formattedDate);
+                    }
+                    catch (err) {
+                        if (!((err as NodeJS.ErrnoException).code === 'EEXIST')) {
+                            console.error(`FROM MOVE IMAGES SERVICE :::::::: Error creating directory: ${err}`);
+                        }
+                    }
+                }
+                const result = await execAsync(moveFileToRightDir);
```
punov, line 56: If we don't need to use `result`, we can just 
`await execAsync(moveFileToRightDir);`

src/util/motionModel.ts
```diff
@@ -306,7 +364,10 @@ export const getNextGnss = (): Promise<GnssMetadata[][]> => {
     try {
       console.log('Last file is ' + prevGnssFile);
       pathToGpsFile = await getNextGnssName();
+      copyFileToUSB(pathToGpsFile, USB_WRITE_PATH, FILETYPE.GNSS);
```
punov, line 367: @HariHive can we do it only if `pathToGpsFile !== prevGnssFile`,
we'll not be copying the same gps file twice then

src/util/motionModel.ts
```diff
@@ -298,6 +307,55 @@ let emptyIterationCounter = 0;
...
+  const parts = gnssFile.split('.');
+  if (parts.length > 1) {
+    const lastPart = parts.pop(); // Remove the last part
+    const replacedString = parts.join('-') + '.' + lastPart;
+    return fileCreationDate + '/' + replacedString;
+  }
+  return '';
+};
+
+const copyFileToUSB = async (fileName: string, sourceFolder: string, fileType: FILETYPE) => {
```
punov, line 321: I'd not pass `sourceFolder` as parameter to this function,
that wouldmake more sense if `copyFileToUSB` function knows where USB folder is located by itself.

src/util/motionModel.ts
```diff
@@ -298,6 +307,55 @@ let emptyIterationCounter = 0;
...
+
+      const fileCreationDate = fileNameForFAT32.split('T')[0];
+
+      const destinationFileName = createFileNameForFAT32(fileNameForFAT32, fileCreationDate);
+
+      if (destinationFileName) {
+        const destinationFilePath = sourceFolder + '/' + destinationFileName;
+
+        try {
+          await fs.mkdirSync(path.join(sourceFolder, fileCreationDate));
```
punov, line 343: @HariHive do you mind creating a sub-folder for every file type?
so it will be 
/2023-10-10
  /gps
  /imu
  
I think it will be slightly more clean to put it in subfolders, otherwise the day folder itself will still be looking pretty messy.

punov (changes requested):

punov (approved):

PR #108, opened 2023-10-13
Check if directory exists for GNSS and IMU then copy data

HariHive:


punov:
Closed for now,
since there's no more GPS/IMU JSON logs on latest. Need to revisit with SQLite solution

PR #110, opened 2023-10-17
Bugfix/donot allow upload of firmware when it is already there

HariHive:


src/routes/upload.ts
```diff
@@ -1,20 +1,29 @@
...
+    const fileSize = req?.query?.fileSize;
+    console.log(`From upload route: Uploaded File size: ${fileSize}`);
     req.busboy.on('file', (fieldname, file, data) => {
       const filename = data && data.filename ? data.filename : 'upload.raucb';
       console.log(`Upload of '${filename}' started`);
       // Create a write stream of the new file
       try {
         try {
           if (existsSync(UPLOAD_PATH + filename)) {
+            const existingFileSize = parseInt(execSync(`ls -l ${UPLOAD_PATH+filename} | awk '{print $5}'`));
```
punov, line 20: @HariHive this one looks slightly tricky,
let's maybe do it via promisified stat() function? Should look simpler and more readable

src/routes/upload.ts
```diff
@@ -1,20 +1,29 @@
...
+            console.log(`File size is ${existingFileSize} bytes`);
+            if (fileSize && typeof fileSize === 'string' && parseInt(fileSize) === existingFileSize) {
```
punov, line 22: I think slightly better API would be to pass `size` as number via POST body (req.body.size) or via data fields that you use above (data.size),
this way we will be able to pass the number, and no need to workaround it with strings.

punov (changes requested):

src/routes/upload.ts
```diff
@@ -1,36 +1,65 @@
...
+      console.log(`Size of upload file is ${file}`);
+      fileSize = file;
+    });
     req.busboy.on('file', (fieldname, file, data) => {
       const filename = data && data.filename ? data.filename : 'upload.raucb';
       console.log(`Upload of '${filename}' started`);
       // Create a write stream of the new file
       try {
         try {
           if (existsSync(UPLOAD_PATH + filename)) {
```
punov, line 23: `UPLOAD_PATH + filename` repeated about 5 times,
so it will make sense for me to do:
`const uploadFilePath = UPLOAD_PATH + filename`

src/routes/upload.ts
```diff
@@ -1,36 +1,65 @@
...
+            // Pipe it trough
+            file.pipe(fstream);
+
+            // On finish of the upload
+            fstream.on('close', () => {
+              console.log(`Upload of '${filename}' finished`);
+              res.json({
+                output: 'done',
+              });
+            });
```
punov, line 57: Since the same write stream block is used twice here, I think we can make it cleaner in a way:
```
if (exists) {
  if (same file size) {
    res.json();
    return; // it will return here and will not get to uploading code part
  } else {
    rmSync()
  }
}
// and then uploading code:
const fstream = ..
```

punov (changes requested):

src/routes/upload.ts
HariHive: updated

src/routes/upload.ts
HariHive: updated

punov (approved):

punov:
Closed as outdated

Issue #113, opened 2023-10-23
Read SQLite data frameKm should be a km of data


jayalberts:
logic should not be subsequent rules (e.g. GNSS 3 points)

include stopping condition for bad IMU

jayalberts:
discovered issues chunking framekms and working on solution. Ability to finish today depends on data validation 3.6.

jayalberts:
working on slight refactor for persistent behaviors in SQLite instead of folder system

jayalberts:
Some topics to discuss with Evan related to motion model math -- send in slack. Create PR for Evan, sooner the better.

punov:
In review: https://github.com/Hivemapper/odc-api/pull/117
Moving to testing tomorrow, including all the fixes needed

jayalberts:
finished architectural skeleton and shared with Evan, focus today is on testing and discovering failure modes. Adding safety checks.

PR #117, opened 2023-11-07
Packaging a kilometer of data into FrameKM #113

punov:


src/sqlite/framekm.ts
```diff
@@ -0,0 +1,196 @@
...
+
+let metersTrimmed = 0;
+
+export const addFramesToFrameKm = async (rows: FramesMetadata[], tableName = 'framekm'): Promise<void> => {
+  const {
+    isTripTrimmingEnabled,
+    TrimDistance
+  } = getConfig();
+
+  if (isTripTrimmingEnabled && metersTrimmed < TrimDistance) {
```
punov, line 90: Trip trimming logic lives here for now.
Every time we try to insert new frames into FrameKM table,
if trip trimming is enabled and it was not yet trimmed enough during this session,
it will push those frames into `prev_framekm` table instead of `framekm` table.

src/services/motionModel.ts
```diff
@@ -3,21 +3,48 @@ import {
...
 const execute = async () => {
   try {
+    const {
+      isTripTrimmingEnabled,
+      TrimDistance,
+      DX
+    } = getConfig();
+
+    if (isTripTrimmingEnabled && !endTripTrimmed) {
+      const data = await getFrameKmMetadata();
```
punov, line 36: Start trip trimming:

Executes on start, checks the contents of FrameKM table (leftovers of previous drive).
Trims last N records depending on TrimDistance (configurable),
packs and cleans up the table and frame folders.

PR #137, opened 2023-12-12
Add endpoint for getting sensor telemetry from the camera.

mazar-farran:
This endpoint delivers IMU and GNSS data.
This endpoint will be periodically polled by the mobile app.

Ticket: Hivemapper/mobile#1440

- [x] semantic version was incremented inside `src/config/index.ts`

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -1,5 +1,5 @@
 export * from './hdc';
-export const API_VERSION = '3.8.1';
+export const API_VERSION = '3.9.0';
 
 export const isDev = () => {
   return false;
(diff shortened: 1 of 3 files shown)


src/sqlite/gnss.ts
```diff
@@ -3,11 +3,11 @@ import { db } from './index';
 import { convertTimestampToDbFormat } from 'util/index';
 
 export const fetchGnssLogsByTime  = async (from: number, to?: number): Promise<GnssRecord[]> => {
-    let query = `SELECT * FROM gnss WHERE time > ?`;
+    let query = `SELECT * FROM gnss WHERE system_time > ?`;
```
punov, line 6: @mazar-farran Motion model expects this part to select based on `time`,
as a reliable metrics of a real current time, so let's bring it back to `time`, if you don't mind!

punov (changes requested):

src/sqlite/gnss.ts
mazar-farran: Done, changed back to `time`

mazar-farran:
Changed to 3.8 branch.

punov (approved):

PR #141, opened 2023-12-28
Feature/check acl before downloading framekms

HariHive:


src/routes/metadata.ts
```diff
@@ -7,12 +7,14 @@ import { setMostRecentPing } from 'services/heartBeat';
...
 import { promisify } from 'util';
+import { ensureAclPassed } from './middleware/acl';
 
 const MAX_RESPONSE_SIZE = 10000;
 const router = Router();
 
-router.get('/', async (req: Request, res: Response) => {
+router.get('/', ensureAclPassed, async (req: Request, res: Response) => {
   try {
+    console.log('[METADATA] Control reached to metadata route fetching the data');
```
punov, line 17: If you don't mind, 
let's remove this one just in case,
App calls this pretty frequently, it will make the dashcam logs very messy

src/routes/middleware/acl.ts
```diff
@@ -0,0 +1,93 @@
...
+import { ACL_FILES_PATH, ACL_TOOL_PATH } from '../../config';
+import { promisify } from 'util';
+import e, { NextFunction, Request, Response, Router } from 'express';
+import {
+  fromString as fromHexString,
+} from 'hex-array';
+const execPromise = promisify(exec);
+
+export async function ensureAclPassed(req: Request, res: Response, next: NextFunction) {
+  console.log("[ACL] Control reached middleware ensureAclPassed req.headers.cookie value", req.headers.cookie);
```
punov, line 11: same here, thank you!

src/routes/middleware/acl.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  if (req.headers.cookie) {
```
punov, line 13: If client doesn't provide the cookie, it will allow to read all the files.
It makes a good backward compatibility with old Apps,
but I think it will be better to always expect the cookie — otherwise to not allow this at all.

src/routes/middleware/acl.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    const accessControlList = await getAccessControlListFromCamera();
```
punov, line 15: I'd move this check to the very top.
If there's no ACL on camera, then go next() immediately, no need to restrict anything

src/routes/middleware/acl.ts
punov: But maybe... Let's keep the logs while we are testing it.
Just let's make sure it will not be too massive: having 6-7 lines of logs every 5 seconds will blow up the log file

punov (changes requested):

punov (approved):

PR #146, opened 2024-01-04
Sensor telemetry improvement

mazar-farran:
Add /sensorquery endpoint which supports `since` and `until`.  Returns all sensor data that falls within these timestamps.

Ticket: Hivemapper/mobile#1551

### DIFF
--- a/src/routes/index.ts
+++ b/src/routes/index.ts
@@ -254,8 +254,6 @@ router.post('/cmd/sync', async (req, res) => {
   }
 });
 
-
-
 router.get('/sensordata/:since', async (req: Request, res: Response) => {
   let since: number;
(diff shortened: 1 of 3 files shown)


src/routes/index.ts
```diff
@@ -254,8 +254,6 @@ router.post('/cmd/sync', async (req, res) => {
   }
 });
 
-
-
 router.get('/sensordata/:since', async (req: Request, res: Response) => {
```
mazar-farran, line 257: This endpoint is no longer needed, but I don't want to break sensor telemetry on the old mobile app firmware.  What should we do with this?

punov (approved):

PR #158, opened 2024-01-26
Change `ifTimeSet` to `isTimeSet`.

mazar-farran:
Very minor change. `ifTimeSet` is not usual for a predicate function, usually "if" stands for "interface".

I don't think this needs a ticket or semantic version bump.

### DIFF
--- a/src/util/lock.ts
+++ b/src/util/lock.ts
@@ -2,14 +2,9 @@ import { exec, ExecException } from 'child_process';
 
 export const DEFAULT_TIME = 1704096000000; // 2024-01-01, dashcam default time is less than this date. So once the date is bigger, we know that system time is set
 let lockTime = 0;
-let isTimeSet = false;
 
-export const ifTimeSet = () => {
-  if (isTimeSet) {
-    return true;
-  }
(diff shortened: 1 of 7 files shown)


src/util/lock.ts
```diff
@@ -2,14 +2,14 @@ import { exec, ExecException } from 'child_process';
 
 export const DEFAULT_TIME = 1704844800000; // 2024-01-10, dashcam default time is less than this date. So once the date is bigger, we know that system time is set
 let lockTime = 0;
-let isTimeSet = false;
+let timeSet = false;
 
-export const ifTimeSet = () => {
-  if (isTimeSet) {
+export const isTimeSet = () => {
+  if (timeSet) {
```
mazar-farran, line 8: Is this caching necessary? Seems like overkill to make this a stateful function just to avoid calling Date.now().

punov (approved):

src/util/lock.ts
punov: @mazar-farran that was needed before, no need for this anymore, thank you!

punov (approved):

PR #159, opened 2024-01-29
Add DashcamUSBState event

mazar-farran:
Hivemapper/docs#346

Add DashcamUSBState event that lets us know when usb is connected or disconnected.

The implementation simply polls the the mount point to check if it's present. It only sends an event if the state changed.

Screenshot 2024-01-29 at 1 29 15 PM

- [x] semantic version was incremented inside `src/config/index.ts`


mazar-farran:
Changed interval to 15 seconds and added usb info to DashcamPackedFrameKm as requested.

punov (approved):

PR #165, opened 2024-02-14
Added anonymous ID to dashcam and metadata files

HariHive:
https://github.com/Hivemapper/docs/issues/361


mazar-farran:
If I'm understanding correctly, with this change, we are generating an anonymous ID on odc-api initialization. Then we store it in a new table `deviceInfo` for anonymous ID's.  This anonymous ID is sent inside of FrameKM metadata.

Just so I understand better, what is the purpose of persisting the anonymous ID if we are creating a new anonymous ID on odc-api start?  Is another process using this ID?

HariHive:
> If I'm understanding correctly, with this change, we are generating an anonymous ID on odc-api initialization. Then we store it in a new table `deviceInfo` for anonymous ID's. This anonymous ID is sent inside of FrameKM metadata.
> 
> Just so I understand better, what is the purpose of persisting the anonymous ID if we are creating a new anonymous ID on odc-api start? Is another process using this ID?

@mazar-farran  We create an Anonymous ID when not present, i.e create only once and later we use it again and again. 
That is why I have **Insert or Ignore** in place for adding anonymous ID into deviceInfo table.
I think we eventually want to save all deviceInfo like ssid, serial number and other values also into DeviceInfo table along with Anonymous ID, hence we are persisting the data

mazar-farran:
Ok, I see, I missed the `INSERT OR IGNORE`. Then LGTM 

please also add the ticket to the PR description

mazar-farran (approved):

src/config/hdc.ts
```diff
@@ -11,7 +11,7 @@ export const FRAMEKM_ROOT_FOLDER = '/mnt/data/framekm';
 export const UNPROCESSED_FRAMEKM_ROOT_FOLDER = '/mnt/data/unprocessed_framekm';
 export const RAW_DATA_ROOT_FOLDER = '/mnt/data/raw';
 export const STREAM_REQUEST_FOLDER = '/mnt/data/request';
-export const DB_PATH = '/mnt/data/data-logger.v1.4.2.db';
+export const DB_PATH = '/mnt/data/data-logger.v1.4.3.db';
```
punov, line 14: No need to uptick DB version,
new table creation is a safe change for other data!

src/util/motionModel/packaging.ts
```diff
@@ -175,6 +177,7 @@ export const packMetadata = async (
         keyframeDistance: getConfig().DX,
         resolution: '2k',
         version: '1.8',
+        deviceAnonymousId: deviceAnonymousId,
```
punov, line 180: we can also name it smth simple like `deviceId`,
since it will be a name to go full stack,
but let me know if you were asked to make it like this

src/util/motionModel/packaging.ts
```diff
@@ -175,6 +177,7 @@ export const packMetadata = async (
...
       },
```
punov, line 181: Also callout:
IIRC, that can break the App download/upload since App is pushing all the metadata JSON fields into its own tables too.
So releasing prior the App version that supports this new field is required.
Let me know if I'm mistaken

src/config/hdc.ts
HariHive: reverted the change

src/sqlite/deviceInfo.ts
```diff
@@ -0,0 +1,37 @@
+import { generate } from 'shortid';
+import { db, getAsync, runAsync } from './index';
+
+const ANONYMOUS_ID_FIELD = "anonymousId";
+
+export const addDeviceInfo = async (): Promise<void> => {
+    const anonymousId = generate();
+    return new Promise(async (resolve) => {
+        const insertSQL = `
+          INSERT OR IGNORE INTO deviceInfo (
```
punov, line 10: So it will be inserted once, and it will ignore all the other inserts, right?
And only once DB wiped / upticked, new deviceId will be generated and inserted.

Is this logic expected? (just double-checking)

punov (approved):

PR #174, opened 2024-03-06
Add public key retrieval for GNSS Auth.

mazar-farran:
Ticket: https://github.com/Hivemapper/network/issues/2083

Note that this only works for HDC.


src/services/getPublicKeyFromEeprom.ts
mazar-farran, line 1: Where should this go?

punov (approved):

PR #175, opened 2024-03-07
Add deviceId to sensorquery.

mazar-farran:
Ticket: https://github.com/Hivemapper/hdcs_firmware/issues/66

Also removes /sensordata because that is no longer being used. This should cause no issues.  If there are older app versions out there that still use this, they will simply fail to retrieve sensor data


mazar-farran:
Tested locally after rebasing, seems to work

punov (approved):

PR #182, opened 2024-03-28
Add deviceId to `/info` api.

mazar-farran:
Ticket: https://github.com/Hivemapper/mobile/issues/1732

This is needed to distinguish between data from different sessions.


mazar-farran:
Not needed, we can just use the deviceId in the sensor data

PR #204, opened 2024-05-21
Add workflow for testing odc-api end to end.

mazar-farran:
Ticket: https://github.com/Hivemapper/docs/issues/404


end-to-end-test/test_finish.sh
```diff
@@ -0,0 +1,51 @@
+METADATA_PATH=./mnt/data/metadata
+FRAMEKM_PATH=./mnt/data/framekm
+REFERENCE_PATH=./reference
+
+FAILURE=0
+
+echo "Testing odc-api"
+echo "Checking for files in framekm folder"
+
+if [[ `ls $FRAMEKM_PATH | wc -l` -eq 0 ]]
```
punov, line 10: Since the overall project is an API server too,
it would be cool to check such things through API endpoints, like:
GET framekm/
GET metadata/

Then we also combine the check for packaged data with a check of API serving the files properly

mazar-farran: I did not consider that we could use these API endpoints to test framekm packaging.  In that case we could probably make use of more sophisticated test frameworks than what we're doing

punov (approved):

mazar-farran:
Did a sanity test without the end-to-end testing enabled. I was able to get a gps lock outside and read sensor data. `info` api worked.

PR #208, opened 2024-05-23
[Feature] IMU stationary detection

### LINKED ISSUE #202: Use Sensor Fusion approach to detect stationary and filter out at motion model (i.e., needs to be moving to add to the model)
We want to filter out data where we have "dancing position" when car has GPS but is stationary

roger-berman:
Here is the PR that surfaces the stationary vs not stationary signal to the motion controller


punov:
Closed since fully revisiting by another approach

Issue #205, opened 2024-05-23
ODC-API E2E Test -- Add more test coverage


mazar-farran:

Masaya-RT:
Working on generalizing what he did to work with multiple tests.
Mostly done but uncovered issue: not comparing correct things
- Some issue with comp where slightly inequal numbers fails whole test

PR #209, opened 2024-05-24
Generalize ODC-API end-to-end testing to work for multiple tests.

mazar-farran:
Ticket: https://github.com/Hivemapper/odc-api/issues/205

Usage covered in https://www.notion.so/WIP-End-to-End-Integration-Testing-for-ODC-API-7e9e1929f47a430d9b3daa07a330978a


mazar-farran:
Current state: It works and most of the code was generalized to support multiple tests.

Downloading the dbs is still too difficult to do for multiple tests. We need to switch to AWS CLI with a restricted IAM role.  This way we can download all tests and test data with a single command.

punov (approved):

Issue #215, opened 2024-07-02
Move firmware version from odc-api to dashcam firmware.


mazar-farran:
Updating odc-api just to update the firmware version is annoying and not necessary. We should move firmware version to somewhere in the firmware itself (e.g. a file) and have odc-api read that.

Linear ticket: https://linear.app/hivemapper/issue/IMCO-50/move-firmware-version-from-odc-api-to-dashcam-firmware

mazar-farran:
Done

PR #216, opened 2024-07-02
Pass odc version as an environment variable.

mazar-farran:
Ticket: https://github.com/Hivemapper/odc-api/issues/215

This allows us to specify the odc version via the service file.


mazar-farran:
Verified this works on HDC

punov (approved):

PR #222, opened 2024-07-29
Stationary Feature FrameKM Integration

roger-berman:
PR integrates the stationary feature into odc-api.
Shows frames tagged as stationary in packed_framekms table.

### DIFF
--- a/src/sqlite/common.ts
+++ b/src/sqlite/common.ts
@@ -42,7 +42,7 @@ export const querySensorData = async (
         const imuSince = gnss[0].system_time;
         const imuUntil = gnss[gnss.length - 1].system_time;
         const session = gnss[0].session;
-        const imu = await fetchProcessedImuLogsByTime(imuSince, imuUntil, session); // What happens if the data is partially missing? ex. imu data is present for half the queried time period.
+        const imu = await fetchProcessedImuLogsByTime(imuSince, imuUntil, session); 
         await sleep(2000); // let frame buffer to fill up if needed
         const images = await getFramesFromFS(imuSince, imuUntil);
         let magnetometer: MagnetometerRecord[] = [];
--- a/src/sqlite/common.ts
+++ b/src/sqlite/common.ts
@@ -22,14 +22,14 @@ let biggestGnssTimeDelta = 0;
 let biggestTimeFixDelta = 0;
 
 export const querySensorData = async (
-  since: number, until?: number, ignoreMagnetometer?: boolean
+  since: number, until?: number, ignoreMagnetometer?: boolean, requester?: string,
 ): Promise<{ gnss: GnssRecord[]; imu: ImuRecord[]; images: IImage[]; magnetometer: MagnetometerRecord[] }> => {
   try {
     if (!since) {
       return { gnss: [], imu: [], images: [], magnetometer: [] };
     }
     const start = Date.now();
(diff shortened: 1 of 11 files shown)


src/sqlite/common.ts
```diff
@@ -31,18 +31,18 @@ export const querySensorData = async (
     const start = Date.now();
     console.log('Getting sensor data for: ', new Date(since));
 
-    // Restricting the GNSS query to 2 min max, to prevent accidental overloads.
+    // Restricting the GNSS query to 2 min max, to prevent accidental overloads. // What is meant by accidental overloads? Important given IMU data is now running at 100 Hz.
```
punov, line 34: Just to prevent accidentally querying the entire table

src/sqlite/common.ts
```diff
@@ -31,18 +31,18 @@ export const querySensorData = async (
...
     }
 
-    const gnss = (await fetchGnssLogsByTime(since, until)).filter(g => g);
+    const gnss = (await fetchProcessedGnssLogsByTime(since, until)).filter(g => g);
     if (gnss.length) {
         const imuSince = gnss[0].system_time;
         const imuUntil = gnss[gnss.length - 1].system_time;
         const session = gnss[0].session;
-        const imu = await fetchImuLogsByTime(imuSince, imuUntil, session);
+        const imu = await fetchProcessedImuLogsByTime(imuSince, imuUntil, session); // What happens if the data is partially missing? ex. imu data is present for half the queried time period.
```
punov, line 45: Ideally both tables syncronised on the latest system time,
otherwise it can start creating holes in FrameKMs

src/sqlite/gnss.ts
```diff
@@ -31,37 +31,69 @@ export const fetchGnssLogsByTime  = async (from: number, to?: number): Promise<G
...
+
+export const fetchLastNProcessedGnssRecords = async (n: number): Promise<GnssRecord[]> => {
+    const query = `SELECT * FROM gnss_processed ORDER BY id DESC LIMIT ?`;
+    return queryDB<GnssRecord>(query, [n]);
+}
+
+export const fetchLastProcessedGnssRecord = async (): Promise<GnssRecord | null> => {
     try {
-        const lastRecords = await fetchLastNGnssRecords(1);
+        const lastRecords = await fetchLastNProcessedGnssRecords(1);
```
punov, line 96: We can leave this one for raw GNSS querying,
this one is to read the last time record from the table to understand the current time — so no need to wait for data to be processed

src/util/motionModel/draftFrameKm.ts
```diff
@@ -28,6 +28,7 @@ export class DraftFrameKm {
   totalDistance = 0;
   highSpeedRecordsInARow = 0;
   lowSpeedRecordsInARow = 0;
+  isStationary = false;
```
punov, line 31: this entire logic in draftFrameKm is not needed here anymore!

src/sqlite/gnss.ts
roger-berman: This was a function that was added, the original is still there and used in the hearBeat.ts file. Git is just being dumb with its diff handling here.

src/util/motionModel/draftFrameKm.ts
roger-berman: kept for later use

PR #224, opened 2024-08-05
Gap Bug Fix

roger-berman:
Allows the motion model to query sensor data by what was last ingested instead of just current time - 60 seconds.

### DIFF
--- a/src/util/motionModel/motionModelController.ts
+++ b/src/util/motionModel/motionModelController.ts
@@ -31,16 +31,8 @@ export async function MotionModelController() {
       await session.checkObjectDetectionService();
     }
 
-    // if sessions last ingested time is 0, query last time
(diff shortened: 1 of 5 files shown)


src/util/motionModel/motionModelController.ts
```diff
@@ -31,8 +31,16 @@ export async function MotionModelController() {
       await session.checkObjectDetectionService();
     }
 
+    // if sessions last ingested time is 0, query last time
+    let queryTime = session.lastIngestedTime;
+    if (session.lastIngestedTime === 0){
```
punov, line 36: Let's keep all this logic inside driveSession,
so motionModel just keeps calling `session.getLastTime()` function

roger-berman: Done

PR #226, opened 2024-08-26
add DB backup functionality

HariHive:

### DIFF
--- a/python/backup.py
+++ b/python/backup.py
@@ -1,0 +1,25 @@
+import sqlite3
+import os
+import sys
+
+if len(sys.argv) != 2:
+    print("Usage: python backup_script.py <source_db>")
+    sys.exit(1)
+source_db = sys.argv[1]
+backup_db = '/data/recording/backup.db'
+
+if os.path.exists(backup_db):
+    os.remove(backup_db)
+    print(f"From backup.py file: Existing backup file {backup_db} deleted.")
+
+source_conn = sqlite3.connect(source_db)
+
+backup_conn = sqlite3.connect(backup_db)
+
+with backup_conn:
+    source_conn.backup(backup_conn, pages=1, progress=None)
+
+source_conn.close()
+backup_conn.close()
+
+print(f"From backup.py file: Backup completed successfully. The backup is stored in {backup_db}")
--- a/python/backup.py
+++ b/python/backup.py
@@ -1,25 +1,32 @@
 import sqlite3
+import argparse
 import os
-import sys
 
-if len(sys.argv) != 2:
-    print("Usage: python backup_script.py <source_db>")
-    sys.exit(1)
-source_db = sys.argv[1]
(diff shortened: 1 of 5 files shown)


python/backup.py
```diff
@@ -0,0 +1,25 @@
+import sqlite3
+import os
+import sys
+
+if len(sys.argv) != 2:
```
roger-berman, line 5: Nit: Can you use argparse instead?
As well as can you put this under a main.
Example of both here: https://github.com/Hivemapper/sensor_fusion/blob/main/sensor_fusion_service/main.py#L269

python/backup.py
```diff
@@ -0,0 +1,25 @@
...
+    print("Usage: python backup_script.py <source_db>")
+    sys.exit(1)
+source_db = sys.argv[1]
+backup_db = '/data/recording/backup.db'
```
roger-berman, line 9: Nit: Can you move this constant to the top and capitalize the entire variable?

src/util/runCommands.ts
```diff
@@ -0,0 +1,24 @@
+import { exec } from 'child_process';
+import { promisify } from 'util';
+
+const asyncExec = promisify(exec);
```
roger-berman, line 4: Does this type of functionality not already exist somewhere int the code? If not then this is totally fine, just wanna make sure we don't have a bunch of code duplication.

src/util/runCommands.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const runSequence = async(cmds: string[]) => {
+    try{
+        for await (const cmd of cmds) {
```
roger-berman, line 8: Does this need to be a for await loop? You already await the asyncExec within the loop?

roger-berman (commented):
Looks good just a few comments

src/util/runCommands.ts
HariHive: I have checked everywhere else looks like promisified exec is used, please lmk if you want me to change it

src/util/runCommands.ts
HariHive: taken this inspiration from src/util/index.ts and src/routes/instrumentation.ts please lmk if you need me to change

mazar-farran: I think the `for await` is not needed, you can change to `for`

mazar-farran:
Would this work?

```
// @ts-ignore
let backup = db.backup('backup.db', (err: any) => {
    backup.step(-1, (err: any) => {
        backup.finish(function(err: any) {
            console.log('Backup finished', err);
        });
    });
});
```

There seems to be a backup api in node-sqlite3 but it's not part of the TypeScript interface.  https://github.com/TryGhost/node-sqlite3/blob/master/lib/sqlite3.js#L105

It's kind of clunky but it might be better than calling python

This worked locally for me on laptop but I haven't tested on HDC

PR #228, opened 2024-09-27
Add snr and exif detections

punov:


src/util/framekm.ts
```diff
@@ -61,6 +81,14 @@ export const concatFrames = async (
     return bytesMap;
   }
 
+  try {
```
mazar-farran, line 84: What does this do? Is it for debugging?

mazar-farran (approved):
LGTM but please add the ticket to the description

src/util/framekm.ts
punov: @mazar-farran yeah, I'll better add some comments here: this is the most performant way to add exif data to batch of images:
1. you first make csv file where you push the path to every image, and all the data you want to push to each of them
2. and then you call exif providing it this csv file as input, then it parses this csv and processes all the images as a single operation
It saves a lot of time

PR #229, opened 2024-10-04
added route for firmware install

HariHive:

### DIFF
--- a/src/routes/firmware.ts
+++ b/src/routes/firmware.ts
@@ -1,0 +1,84 @@
+import { spawn, execSync } from 'child_process';
+import { CAMERA_TYPE } from 'config';
+
+import { Router } from 'express';
+import { CameraType } from 'types';
+const router = Router();
+
+const HDC_ROOT = '/mnt/data/'
+const HDCS_ROOT = '/data/'
+const MENDER_PATH = HDCS_ROOT + 'core.mender'
+const FIP_PATH = HDCS_ROOT + 'fip.bin'
+
+let message = '';
+
+const runSpawn = (cmd: string) => {
+    const child = spawn(cmd, {
+        shell: true
+    });
+    child.stdout.setEncoding('utf8');
+    child.stdout.on('data', (data) => {
+        message = data.toString();
+    });
+
+    child.on('error', (error) => {
+        message = "Error" + error.toString();
+        console.error(`Error in Spawn process for cmd ${cmd} stderr: ${error}`);
+    });
+
+    child.on('close', (code, err) => {
+        if (code !== 0) {
+            message = "Closing spawn with code" + code?.toString() + "Error:" + err?.toString();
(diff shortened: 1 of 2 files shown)


HariHive:
Tested with HDC and HDCS using postman

src/routes/firmware.ts
```diff
@@ -0,0 +1,82 @@
...
+    try {
+        if (CAMERA_TYPE === CameraType.Hdc) {
+            try {
+                execSync(`test -f ${HDC_ROOT + firmwareFile}`, {
+                    encoding: 'utf-8',
+                });
+            } catch (error: unknown) {
+                console.log("Rauc file is not present")
+            }
+            runSpawn(`rauc install ${HDC_ROOT + firmwareFile} && reboot`)
```
punov, line 52: @HariHive @harroyoc 
Have we had a plan to leave the phone a chance to wait for installation to finish and sent "reboot" command?
Or you guys want the dashcam to fully own it?

There's a point of letting the dashcam restart itself, but I'm worried if the phone will be able to properly handle and receive the updated state in this case

src/routes/firmware.ts
```diff
@@ -0,0 +1,82 @@
...
+                console.log("Mender file is not present")
+            }
+            try {
+                execSync(`test -f ${FIP_PATH}  && rm ${FIP_PATH}`, {
+                    encoding: 'utf-8',
+                });
+            } catch (error: unknown) {
+                console.log("Fip file is not present")
+            }
+            runSpawn(`tar -xzf /data/${firmwareFile} -C /data && os-update --install ${MENDER_PATH} && movisoc-fwu -a ${FIP_PATH} && reboot`)
```
punov, line 70: @HariHive 
What we should check here for sure is that if mender for some reason errors out but doesn't break the command chain, we don't want movisoc-fwu to be called.
So we should be 100% sure that failing mender (os-update) will break the command chain for sure

src/routes/firmware.ts
```diff
@@ -0,0 +1,82 @@
...
+            res.json({ output: 'received install command' });
+        }
+    } catch (error: unknown) {
+        res.json({ error });
+    }
+});
+
+router.get('/progress', async (req, res) => {
+    res.json({ message });
```
punov, line 79: @HariHive I deally we give the percentage number back, not just the last message.
So adding the logic to unparse the progress will be good.

progress = 0
....
  on.data -> reads message, parses percentage out of it (when possible), assigns to progress (if successfully parsed)
....
/progress route:
returns progress value

src/routes/firmware.ts
HariHive: Discussed offline, we want app to handle reboot

src/routes/firmware.ts
HariHive: could not check if os-update command breaks, but test with tar command, if tar fails nothing executes:
image

HariHive:
closing as duplicate

PR #235, opened 2024-10-24
Feature/firmware installtion progress indicator

HariHive:

### DIFF
--- a/src/routes/firmware.ts
+++ b/src/routes/firmware.ts
@@ -12,6 +12,7 @@ const FIP_PATH = HDCS_ROOT + 'fip.bin';
 
 let message = 'started';
 let errorSeen = false;
+const SUCCESS_MESSAGE = 'Spwan ran successfully';
 
 const runSpawn = (cmd: string) => {
   const child = spawn(cmd, {
@@ -39,7 +40,7 @@ const runSpawn = (cmd: string) => {
         `Closing spawn due to error for cmd ${cmd} stderr: ${err?.toString()}`,
       );
     } else {
-      message = 'Spwan ran successfully';
+      message = SUCCESS_MESSAGE;
     }
   });
 };
@@ -90,7 +91,7 @@ router.get('/install', async (req, res) => {
 });
 
 router.get('/progress', async (req, res) => {
-  res.json({ isRunning: message !== 'Spwan ran successfully', errorSeen });
+  res.json({ isRunning: message !== SUCCESS_MESSAGE, errorSeen });
 });
 
 export default router;
--- a/src/routes/firmware.ts
+++ b/src/routes/firmware.ts
@@ -6,6 +6,11 @@ const firmwareManager = new FirmwareManager();
 
 router.get('/install', async (req, res) => {
   const firmwareFile = req?.body?.fileName || '';
+  if(!firmwareFile) {
+    res.json({
+        output: "No firmware file provided",
+    })
+  }
   const resp = firmwareManager.installFirmware(firmwareFile);
   res.json(resp);
 });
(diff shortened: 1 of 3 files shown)


HariHive:
image

HariHive:
image

src/routes/firmware.ts
```diff
@@ -0,0 +1,97 @@
...
+        `Closing spawn due to error for cmd ${cmd} stderr: ${err?.toString()}`,
+      );
+    } else {
+      message = SUCCESS_MESSAGE;
+    }
+  });
+};
+
+router.get('/install', async (req, res) => {
+  const firmwareFile = req?.body?.fileName || '';
```
harroyoc, line 50: dont you need to throw here if there is no filename?

src/routes/firmware.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  try {
+    if (CAMERA_TYPE === CameraType.Hdc) {
+      try {
+        execSync(`test -f ${HDC_ROOT + firmwareFile}`, {
+          encoding: 'utf-8',
+        });
+      } catch (error: unknown) {
+        console.log('Rauc file is not present');
```
punov, line 59: Same here: if no rauc, then we don't need to go to `rauc install`

punov (changes requested):

src/routes/firmware.ts
```diff
@@ -0,0 +1,20 @@
...
+const router = Router();
+
+router.get('/install', async (req, res) => {
+  const firmwareFile = req?.body?.fileName || '';
+  const resp = installFirmware(firmwareFile);
+  res.json(resp);
+});
+
+router.get('/progress', async (req, res) => {
+  res.json({ isRunning: message !== SUCCESS_MESSAGE, errorSeen });
```
harroyoc, line 17: let's this endpoint the same way: the logic doesnt live here 🙏 , something like: 
`res.json(getInstallationProgress())`

and created the method in the firmware util

src/routes/firmware.ts
```diff
@@ -0,0 +1,17 @@
...
+const firmwareManager = new FirmwareManager();
+
+router.get('/install', async (req, res) => {
+  const firmwareFile = req?.body?.fileName || '';
+  const resp = firmwareManager.installFirmware(firmwareFile);
+  res.json(resp);
+});
+
+router.get('/progress', async (req, res) => {
+  res.json({ isRunning: firmwareManager.getMessage(), errorSeen: firmwareManager.getErrorSeen() });
```
harroyoc, line 19: sorry but I think it's easier if it return the json. Otherwise if I use somewhere else I would need to duplicate the same code
`getProgressMessage() --> return { isRunning, error }`

lets keep error (not error seen to keep api consistency

harroyoc: so the router response is
`res.json(getInstallationProgress());`

harroyoc: and wherever else I use I can just call `getInstallationProgress` and we have the same api response type (on the socket for example)

src/routes/firmware.ts
```diff
@@ -0,0 +1,22 @@
...
+import FirmwareManager from 'util/firmware';
+const router = Router();
+
+const firmwareManager = new FirmwareManager();
+
+router.get('/install', async (req, res) => {
+  const firmwareFile = req?.body?.fileName || '';
+  if(!firmwareFile) {
+    res.json({
+        output: "No firmware file provided",
```
harroyoc, line 11: return `{ error: "No firmware file provided" }`
to keep consistency with the installation method

punov (approved):

mazar-farran (approved):

PR #236, opened 2024-10-31
add configurable ppz support

HariHive:


HariHive:
This pr helps to configure radius per privacy zone

src/util/privacy.ts
```diff
@@ -4,24 +4,41 @@ import { getConfig } from 'sqlite/config';
...
   if (!lat || !lon) {
     return false;
   }
   const { privacyRadius } = await getConfig('privacyRadius');
   const { x, y } = latLonToWebMercator(lat, lon);
   const radius = privacyRadius || DEFAULT_RADIUS;
-  return privateZones ? !!privateZones.within(x, y, radius * 1.25).length : false;
-}
+  return privateZones
+    ? !!privateZones.within(x, y, radius * 1.25).length
```
punov, line 20: @HariHive 
So new radius for every privacy zone is not getting there, if I get it right?
Meaning `isPrivateLocation` will use the default radius instead of using the one for particular private zone

punov (changes requested):

PR #239, opened 2024-11-02
new acl routes

HariHive:


HariHive:
This PR has 3 new routes:
1. To create acl if it does not exist
2. To clear acl if it exists
3. Find if camera is locked to logged in user

punov (approved):

PR #242, opened 2024-11-08
Add LTE upload pipeline to Bee

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-337/enable-lte-data-upload

Adds new endpoint /auth for authentication
Adds lteEnabled to /info
Adds service for checking for framekms to upload, and then uploading them.


punov (approved):

src/routes/metadata.ts
```diff
@@ -7,10 +7,16 @@ import { getMetadataFiles, getNumFramesFromChunkName } from 'util/framekm';
 import { join } from 'path';
 import { promisify } from 'util';
 import { ensureAclPassed } from './middleware/acl';
+import { getConfig } from 'sqlite/config';
 
 const router = Router();
 
 router.get('/', ensureAclPassed, async (req: Request, res: Response) => {
+  if (await getConfig('isLTEUploadEnabled')) {
```
harroyoc, line 15: I think this validation should be made at getMetadataFiles level, as Im calling the same when Im sending the metadata through the socket (search who is using calling `getMetadataFiles()`)

mazar-farran: Addressed in https://github.com/Hivemapper/odc-api/pull/245

PR #246, opened 2024-11-14
Fix LTE Upload issues from drives

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-337/enable-lte-data-upload

Restart lte if frequent dns failures.
Correctly check error message of FrameKM Request response.


derekhive (approved):
drive tested 11/14 w/ John

PR #250, opened 2024-11-22
Fix framekm upload version

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-381/lte-upload-landmark-bug


mazar-farran:
Tested and verified that it uploads a framekm

src/services/getFrameKmsService.ts
```diff
@@ -35,7 +35,7 @@ const promiseExec = promisify(exec);
 
 
 async function postFrameKmRequest(framekm_name: string) {
-    const url = `${BASE_URL}/framekm/request/1.8`;
+    const url = `${BASE_URL}/framekm/request/${FRAMEKM_VERSION}`;
```
punov, line 38: Very small callout:
I'd probably better use FrameKM version from metadata, cause of a single scenario:
- framekm collected with version N but not uploaded yet
- the firmware updated to N+1 version
- postFrameKmRequest will then try to upload FrameKM vN using URL for N+1

punov (approved):

src/services/getFrameKmsService.ts
mazar-farran: Changed to check the metadata version

PR #255, opened 2024-12-02
Replace node persist with sqlite

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-394/replace-node-persist-with-sqlite


src/sqlite/index.ts
```diff
@@ -281,6 +282,20 @@ export const createFrameTable = async (db?: Database): Promise<void> => {
   }
 };
 
+const createLTEUploadTable = async (db?: Database): Promise<void> => {
+  const createTableSQL = `
+    CREATE TABLE IF NOT EXISTS lte_upload (
+    key TEXT PRIMARY KEY NOT NULL,
+    value TEXT NOT NULL
```
punov, line 289: Mobile App analogue of this table contains much more data about the retries and statuses for helpful tracking:
https://github.com/Hivemapper/mobile/blob/feature/androidBgDownloads/utils/datastore/frameKm.ts
It's not urgent to add it now, but just commenting out that there could be a utility in having that being tracked as part of the table state

punov (approved):

src/sqlite/index.ts
mazar-farran: Ok, will add these later if Henry requests them

PR #256, opened 2024-12-02
Improve LTE reconnect logic

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-397/improve-lte-reconnect-logic

If ping fails 5 times, restart lte.  Wait up to 60 seconds for it to come back up.


punov (approved):

mazar-farran:
In the future, I think we should decouple LTE health from the upload pipeline (i.e. have a separate service)

PR #257, opened 2024-12-03
Add endpoint for log paths.

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-396/fetch-lte-logs-with-the-app
```
http://192.168.0.10:5000/api/1/logging/files

{"files":["lte-status.log","odc-api.log","hivemapper-data-logger.log","redis-handler.log","map-ai.log"]}
```


src/routes/logs.ts
```diff
@@ -0,0 +1,15 @@
+
+import { Router } from 'express';
+
+const router = Router();
+
+router.get('/paths', (req, res) => {
+  res.json({
+    paths: [
+      '/data/recording/lte-status.log',
+      '/data/recording/odc-api.log',
```
harroyoc, line 10: i think we are missing one here map-ai. Can you check with @derekhive or @punov

mazar-farran: I think that's everything, please note that `/data/recording/hivemapper-data-logger.log` and `/data/recording/redis-handler.log` are not present yet

punov: Yes, `/data/recording/map-ai.log` needs to be there!

harroyoc: also @mazar-farran just pass the file name like "odc-api.log" because I call it from /public/odc-api.log

punov (approved):

PR #259, opened 2024-12-04
Enable LTE uploads if internet connection is established.

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-374/lte-upload-edge-cases

Previously flag gated, now we switch to LTE if we are ever able to ping.

Also changed lteEnabled in `/info` to match this status


punov (approved):

mazar-farran:
Tested on device with and without SIM, and it changed modes as expected after reboot.

Note that the device still needs to be rebooted after SIM is inserted. `AT+CPIN?` will return an error unless there is a SIM on power up.

PR #260, opened 2024-12-06
Feature/installation percentage

harroyoc:


HariHive:
We are still unaware of progress of Unzip command right ?

src/util/firmware.ts
```diff
@@ -44,27 +47,53 @@ class FirmwareManager {
...
   private runSpawn(cmd: string) {
     this.updateStatus(InstallStatus.Installing, 'Installation in progress...');
 
     const child = spawn(cmd, { shell: true });
+    let output = '';
 
     child.stdout.setEncoding('utf8');
     child.stdout.on('data', data => {
-      this.updateStatus(InstallStatus.Installing, data.toString());
+      output += data.toString();
```
HariHive, line 89: Please correct me if I am missing something here:
we can also do: output = data.toString() right ?
as every line has percentage value
this way we do not have to loop through whole output and just parse last line

HariHive: if we are concerned that not every line has percentage, then we can probably store it in a var like "lastKnownPercent" value ?

harroyoc: `output = data.toString()`
No , that would output only the streaming portion received like "...."

harroyoc: `We are still unaware of progress of Unzip command right ?`
Im ok with that, it's just a few seconds

HariHive: > if we are concerned that not every line has percentage, then we can probably store it in a var like "lastKnownPercent" value ?
so this is still a possibility right ?

harroyoc: not sure I understand this:
`if we are concerned that not every line has percentage, then we can probably store it in a var like "lastKnownPercent" value ?`

harroyoc: I prefer not to store any value because we would need to "clean it"

harroyoc: so the regex will give me the last percentage found... so if we are on 27%... and for the next 10 lines there is no percentage we still send 27

src/routes/firmware.ts
```diff
@@ -20,4 +22,9 @@ router.get('/progress', async (req, res) => {
   res.json(firmwareManager.getProgress());
 });
 
+router.post('/complete', async (req, res) => {
+  exec('reboot'); // Restart the dashcam to apply the new firmware
+  res.json({});
```
punov, line 27: It's funny it will most probably not even get to this line, be ready for that!

punov (approved):

src/routes/firmware.ts
harroyoc: true

PR #264, opened 2024-12-10
Fallback to Wi-Fi if LTE fails too much

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-409/collected-data-wasnt-uploaded-from-device-with-no-lte

Permanently drop to Wi-Fi after 5 minutes of failure.


punov (approved):

mazar-farran:
Tested locally and it falls back to Wi-Fi if there's no LTE. The positive case (maintaining LTE after a temporary break) would be difficult to test in office.

PR #265, opened 2024-12-11
support plugins

HariHive:


mazar-farran (approved):
LGTM, could you link to a ticket in the description?

src/sqlite/plugins.ts
```diff
@@ -0,0 +1,73 @@
+import { startPlugin, stopPlugin } from 'services/pluginService';
+import { getAsync, runAsync } from 'sqlite';
+
+export const ENABLE_PLUGIN = 'systemctl enable ';
+export const START_PLUGIN = 'systemctl start ';
+export const DISABLE_PLUGIN = 'systemctl disable ';
+export const STOP_PLUGIN = 'systemctl stop ';
```
punov, line 7: We can move those ones to service itself,
the commands not related much to sqlite!

src/routes/plugin.ts
```diff
@@ -0,0 +1,22 @@
...
+
+router.get('/', async (req, res) => {
+  res.json({
+    status: "success",
+  });
+});
+
+router.post('/setPlugin', async (req: Request, res: Response) => {
+  const plugin  =  req.body.plugin;
+  const state = req.body.state ?? PluginState.DISABLED;
```
punov, line 14: Since we have only two states (enable / disable), I'd prefer organising it as two routes:
/enable/:name
/disable/name
so there's no potential mistake of providing the wrong state.
But your approach though can provide us some versatility in perspective if we want those states to be more complex.

all the rest looks good to me, @HariHive , thank you!

mazar-farran:
Should this have a way to register/unregister plugins? The way it is right now, we can enable/disable any service.  

Also note that we're passing an arbitrary string and running it in shell. Right now there are easier ways to get into the system. But in the future, if we lock things down, this will be a way for a malicious user to execute arbitrary commands

HariHive:
Also includes Mazar's changes from : https://github.com/Hivemapper/odc-api/pull/270/files

PR #301, opened 2025-01-09
Beekeeper device auth

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/HIVE-421/beekeeper-authentication-with-hivemapper-network

This is a service that will execute only if beekeeper is Enabled
It will check if a stored cookie is present and valid. If not it will attempt to authenticate with hivemapper using gnss (endpoint of this is still in progress)

Context: https://hivemapper.slack.com/archives/C087F1ACVPH/p1736379307105499


punov (approved):

mazar-farran:
Working, I was able to authenticate with P1-008

punov (approved):

PR #307, opened 2025-01-15
Landmark mode

punov:
- [ ] Linear ticket: https://linear.app/hivemapper/issue/HDM-20/landmark-mode-support

Dashcam packs two FrameKM versions: one is full and one is for frame of interest (short), for now containing only frames with landmarks and its neighbors.
- That involves new FrameKM_Version to interact with server
- Server expects to know the bytesize of both, full and short FrameKMs
- Server will respond with type of which FrameKM is needed: full or short
- Dashcam will upload the one needed


src/util/motionModel/packaging.ts
```diff
@@ -516,13 +546,17 @@ export const packMetadata = async (
         acc_z: m.acc_z,
         gyro_x: m.gyro_x,
         gyro_y: m.gyro_y,
-        gyro_z: m.gyro_z
+        gyro_z: m.gyro_z,
```
nhyman, line 549: do you know how these get reordered? On the backend they are parsed in this order:

```        case 'lat':
          return 3;
        case 'lon':
          return 2;
        case 'alt':
          return 5;
        case 'speed':
          return 4;
        case 't':
          return 1;
        case 'satellites':
          return 6;
        case 'dilution':
          return 7;
        case 'bytes':
          return 0;
        case 'acc_x':
          return 8;
        case 'acc_y':
          return 9;
        case 'acc_z':
          return 10;
        case 'gyro_x':
          return 11;
        case 'gyro_y':
          return 12;
        case 'gyro_z':
          return 13;
        case 'xdop':
          return 14;
        case 'ydop':
          return 15;
        case 'pdop':
          return 16;
        case 'hdop':
          return 17;
        case 'vdop':
          return 18;
        case 'tdop':
          return 19;
        case 'gdop':
          return 20;
        case 'eph':
          return 21;
        case 'cno':
          return 22;
        case 'hasImage':
          return 23;
```

Want to make sure it is in the spot I expect it to be

punov: @nhyman I think, those ones gets ordered on backend during request step, so all should be good

nhyman (approved):

mazar-farran:
https://github.com/Hivemapper/odc-api/pull/307/files#diff-6de2cebb611c6e6e6046fac38cecedc9bb5fed82b9eeb20e6bf48a95b7e8c9fcL359-L362
https://github.com/Hivemapper/odc-api/pull/307/files#diff-6de2cebb611c6e6e6046fac38cecedc9bb5fed82b9eeb20e6bf48a95b7e8c9fcL392-L395

Should these also delete the new framekm type?

PR #309, opened 2025-01-15
Add /validNavPvt endpoint to retrieve valid gnss entries

HariHive:
this pr gives last n valid GNSS points from nav pvt table, where it checks for valid lat lon and other filters


mazar-farran:
Could you add the endpoint name in the title and a brief description of the endpoint so we can track this better, otherwise LGTM

mazar-farran (approved):

HariHive:
this pr gives last n valid GNSS points from nav pvt table, where it checks for valid lat lon and other filters

PR #312, opened 2025-01-21
Fetch Gnss data for a given session if the session is still valid otherwise fetch lat lon from next session

HariHive:


mazar-farran:
Could you please give this a title and description

src/sqlite/nav_pvt.ts
```diff
@@ -50,4 +50,71 @@ export const fetchLastNValidNavPvtRecords = async (
...
+      AND h_acc_m < 10
+      AND session = ?
+    ORDER BY id;
+  `;
+
+  const db = await getSensorDb();
+
+  // Check if lat/lon exists and fetch data in a single step
+  return new Promise((resolve, reject) => {
+    db.get(
```
mazar-farran, line 95: Wouldn't runAsync work here?

HariHive: sure

HariHive: but getAdync connects to ODC_API db right ?

HariHive: I will need sensors DB

mazar-farran: It looks like you can pass the db to getAsync as the third parameter

HariHive: done

mazar-farran: Sorry, I thought runAsync could be used for both the db all and db get queries, is that not the case?

Will approve, changing the db.all is optional, thanks

mazar-farran (approved):

PR #318, opened 2025-01-23
Add firmware updates over LTE

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-466/bee-firmware-bee-dashcam-direct-lte-updating


src/services/lteFirmwareUpdateService.ts
```diff
@@ -0,0 +1,253 @@
...
+
+function firmwareNameFromUrl(url: string) {
+  const urlParts = url.split('/');
+  return urlParts[urlParts.length - 1];
+}
+
+async function updateIsNewer(url: string) {
+    const urlVersion = versionFromUrl(url);
+    const currentVersion = await getCurrentFirmwareVersion();
+    if (compare(urlVersion, currentVersion, '>')) {
```
HariHive, line 80: This is will not support downgrade of firmwares, if that's a design decision I'm okay with it

HariHive:
Probably not for this PR, but does it make sense to move to ES6 syntax for function declaration ?

src/services/lteFirmwareUpdateService.ts
```diff
@@ -0,0 +1,253 @@
...
+    updateInstalled: false,
+};
+
+export function getLTEFirmwareUpdateStatus() {
+    return lteFirmwareUpdateStatus;
+}
+
+async function getCurrentFirmwareVersion() {
+  const buildInfo = JSON.parse(
+    fs.readFileSync('/etc/build_info.json', 'utf-8'),
```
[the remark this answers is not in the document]
mazar-farran: Not sure I understand, shouldn't `/etc/build_info.json` and `/info` always be in sync?

But I see that this file check is unnecessary, we already have the api version in `API_VERSION`

src/services/lteFirmwareUpdateService.ts
mazar-farran: Yes my understanding was that this was how the app behaves. I can change this to emulate whatever the app is doing

mazar-farran: @harroyoc should I change this to allow rollback like the app?

src/services/lteFirmwareUpdateService.ts
```diff
@@ -0,0 +1,243 @@
...
+          }
+          if (!(await firmwareDownloadedAndValid(url, contentLength, md5sum))) {
+            console.log('Download failed');
+            return;
+          }
+        }
+
+        lteFirmwareUpdateStatus.updateDownloaded = true;
+        console.log('Firmware downloaded and valid, installing');
+        FirmwareManager.getInstance().installFirmware(firmwareNameFromUrl(url));
```
mazar-farran, line 214: @HariHive @harroyoc 

Just want to confirm, is this the code path we use when we do firmware upgrades via app? 

Assuming that the firmware was downloaded, and checksum is passing, would running this be equivalent to upgrading via the app?

HariHive: Assuming that the firmware was downloaded, and checksum is passing, would running this be equivalent to upgrading via the app? => this yes

HariHive (approved):

src/services/lteFirmwareUpdateService.ts
```diff
@@ -0,0 +1,246 @@
...
+            console.log('Error downloading firmware', e);
+            lteFirmwareUpdateStatus.downloadError = String(e);
+          }
+          if (!(await firmwareDownloadedAndValid(url, contentLength, md5sum))) {
+            console.log('Download failed');
+            return;
+          }
+        }
+
+        lteFirmwareUpdateStatus.updateDownloaded = true;
```
punov, line 215: @mazar-farran Could you please share, where this status is used?
Ideally, if we downloaded the firmware and confirmed its valid, we should not download it anymore.
But I'm not sure if this flag is used for it.
Imagine scenario:
- firmware downloaded
- validated
- but failed installation
so we want prevent it from being re-downloaded if not required

src/services/lteFirmwareUpdateService.ts
```diff
@@ -0,0 +1,246 @@
...
+        if (!(await firmwareDownloadedAndValid(url, contentLength, md5sum))) {
+          console.log('New firmware available, downloading');
+          try {
+            await downloadUpdate(url);
+          } catch (e: unknown) {
+            console.log('Error downloading firmware', e);
+            lteFirmwareUpdateStatus.downloadError = String(e);
+          }
+          if (!(await firmwareDownloadedAndValid(url, contentLength, md5sum))) {
+            console.log('Download failed');
```
punov, line 210: Also, if there's a problem of validation,
good idea would be to prevent this service from trying re-downloading the firmware over and over again.
If some instability of LTE prevents the firmware file from being downloaded, we probably don't want to drain the entire plan trying to actively re-download bug file over and over

punov (changes requested):

src/services/lteFirmwareUpdateService.ts
mazar-farran: Everything in lteFirmwareUpdateStatus is only used to surface information to the /lte-debug endpoint.

If firmware is already downloaded, then it will pass the first validation check and not get to the downloadUpdate().

src/services/lteFirmwareUpdateService.ts
mazar-farran: Should I add a limit for download attempts?

mazar-farran: added

punov: @mazar-farran Yes, let's add something reasonable line 2 or 3 tries, 
and also adding a simple instrumentation event to that will help us to proactively discover such cases.

Also, it will make sense to count the download attempts only if LTE was available and healthy.
Probably, ignoring the entire iteration if LTE was not healthy yet?

punov (approved):

PR #319, opened 2025-01-24
Add check to prevent LTE and App from installing firmware at the same time.

HariHive:


HariHive:
https://github.com/user-attachments/assets/e7595a2c-d23a-4253-9b7a-d03f7120ddc6

punov (approved):

PR #323, opened 2025-01-30
Remove LTE health management from odc-api

mazar-farran:
Remove all lte retry logic. It is now the responsibility of lte-init.py to reset the modem

Related to: https://github.com/Hivemapper/hivemapper-hdc-os/pull/136


src/services/getFrameKmsService.ts
```diff
@@ -609,54 +607,15 @@ export const GetFrameKmsService: IService = {
...
 
-      // Temporarily disabled due to issues observed in the field
-      // if (!await isLTEServiceEnabled()) {
-      //     console.log('LTE - Service not enabled');
-      //     isLTEUploadEnabled = false;
-      //     return;
-      // }
       if (!(await isLTEHealthy())) {
         console.log('LTE - Could not ping hivemapper');
         if (lastSuccessfulPing.getTime() < Date.now() - 2 * 60 * 1000) {
```
punov, line 612: Since we don't restart here anymore, do you think that giving the bigger grace period will make sense?
Like 2 minutes -> 5 minutes?

punov (approved):

mazar-farran:
@derekhive FYI with this change we will no longer restart lte with odc-api, so we don't have to worry about this anymore regarding hellbender.

src/services/getFrameKmsService.ts
mazar-farran: done

mazar-farran:
Tested last night and seems to work

PR #324, opened 2025-01-31
clean firmwares after  install

HariHive:
This Pr clears firmware raucb files after successful commit from /data/lte-firmware and  /data folders so techincally after both app installs and lte installs


mazar-farran (approved):
Approved but I want to call out that we're hardcoding these firmware paths in multiple places

e.g. https://github.com/Hivemapper/odc-api/blob/a7d5cd24a8dabe3b8721fe97ff9adb1af288c4b6/src/util/firmware.ts#L7C14-L7C22

Maybe this is for a future improvement but it seems like we need to change the firmware path in a lot of places if we want to update it and Ideally we'd just change one place

HariHive:
let me clean this quickly

HariHive:
cleaned

src/config/index.ts
```diff
@@ -1,4 +1,4 @@
-export * from './hdc';
+export * from './bee';
```
mazar-farran, line 1: change to hdc

mazar-farran (approved):
lgtm

PR #325, opened 2025-02-01
Abort controller

mazar-farran:


mazar-farran:
Worked on a local http server.  If I kill the server, I can replicate the hang. These changes handle the hang correctly and kill the connection after about 1 minute

mazar-farran:
Succeeded by https://github.com/Hivemapper/odc-api/pull/325

PR #326, opened 2025-02-02
Add support for chunked LTE firmware downloads, add AbortController

mazar-farran:
Tickets: https://linear.app/hivemapper/issue/EDGE-480/bee-lte-firmware-chunked-downloads https://linear.app/hivemapper/issue/EDGE-482/investigate-lte-firmware-download-issues-525-to-526
Related to: https://github.com/Hivemapper/hivemapper-hdc-os/pull/142

If a JSON chunk file exists on AWS, we will download the firmware in chunks according to that json.

Each chunk will be md5-checked for integrity.

With this setup it's possible to resume a download from the last chunk.

If no json exists, we download the whole file as before.

This PR also bundles the AbortController PR https://github.com/Hivemapper/odc-api/pull/325. Unfortunately it would be more trouble than it's worth to separate the two now.


mazar-farran:
Verified that chunked download path, missing chunk json path, and abort paths all work

punov (approved):

PR #330, opened 2025-02-07
Update gnss auth to use fleet hash from beekeeper.

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-470/include-fleet-hash-beekeeper-uuid-during-authentication


HariHive:
Just to confirm, this is service that runs periodically right ?

src/services/gnssAuth.ts
```diff
@@ -187,16 +200,24 @@ export const GnssAuthService: IService = {
...
 
         const gnssAuthCredentials = await getGnssAuthCredentials();
+        const fleetHash = await getFleetHash();
 
         if (!gnssAuthCredentials) {
           console.log('GNSS Auth - No gnss auth credentials found');
           return;
+        } else if (!fleetHash) {
+          console.log('GNSS Auth - No fleet hash found');
+          return;
```
mazar-farran, line 210: @HariHive should this field be optional?

mazar-farran: Discussed offline with @jimmdd, all devices with this change should be able to obtain a fleetHash from beekeeper so there is no need to make this field optional

mazar-farran:
@HariHive @jimmdd have all concerns been addressed?

jimmdd (approved):

PR #337, opened 2025-02-13
Add lte redis messages

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-524/update-lte-init-to-use-redis-interface

```http://192.168.0.10:5000/api/1/lte-debug-connectivity-status```
```
{"operator":"Helium","apn":"mob.asm.net","protocol":"IPV4V6","pin_status":"READY","iccid":"8901240154151975410","imeisv":"3503449892706012"}
```

Related to: https://github.com/Hivemapper/hivemapper-hdc-os/pull/159

### CODE REVIEW

File: src/util/lte.ts [hunk 1]
```diff
@@ -118,4 +120,22 @@ export async function uploadLogIfIntervalPassed(filePath: string, intervalMs: nu
         }
     }
     return false;
-}
\ No newline at end of file
+}
+
+export const getLTEServiceStatus = async () => {
+    const operator = await getValueFromRedis(RedisLTEKeys.OPERATOR);
```
punov, line 126: @mazar-farran Does it wait till the key is actually there, or if the key is not there yet — it will return null / undefined?

punov (approved):

src/util/lte.ts
mazar-farran: It will return null if not there

PR #340, opened 2025-02-20
Add ctp app firmware targeting

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-537/match-lte-ota-targeting-to-app-behavior


punov (approved):

mazar-farran:
Verified that the negative case works. Both app targeting and non-app targeting are working

PR #341, opened 2025-02-21
revert pr

HariHive:


derekhive (approved):
Check comparison between before original commit and post reversion: https://github.com/Hivemapper/odc-api/compare/85ff8a1..8d72395

It matches mazar's commit between the two which is the desired behavior: https://github.com/Hivemapper/odc-api/commit/da5cfabf07a32115260bf971661ffa1e2cec296d

PR #342, opened 2025-02-21
Add endpoint to get plugin state

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-548/fw-change-for-plugin-status

Context: https://hivemapper.slack.com/archives/C05JDB2KKH6/p1740173379201649

`http://192.168.0.10:5000/plugin/getPluginState/beekeeper-plugin`

```
{"name":"beekeeper-plugin","state":"enabled"}
```

`http://192.168.0.10:5000/plugin/getPluginState/bad-plugin`

```
{"status":"error","message":"Plugin not found"}
```


src/routes/plugin.ts
```diff
@@ -21,4 +21,22 @@ router.post('/disablePlugin', async (req: Request, res: Response) => {
 
 });
 
+router.get('/getPluginState/:plugin', async (req: Request, res: Response) => {
+  const plugin = req.params.plugin;
+  const state = await getPluginState(plugin);
```
HariHive, line 26: this function `getPluginState` just says if a plugin needs to in enabled or disabled state, but it does not account for the actual state of plugin. Example: getPluginState(beekeeper-plugin) => could return from DB as enabled but when we check `systemctl status beekeeper-plugin` => this may or may not be running

mazar-farran: Is this an expected state (beekeeper enabled but service not running)?  gnss auth, metadata endpoint, etc just checks if beekeeper-plugin database entry is enabled

HariHive: `Is this an expected state (beekeeper enabled but service not running)? gnss auth, metadata endpoint, etc just checks if beekeeper-plugin database entry is enabled` => I have never seen this happen personally, but called it out just in case

mazar-farran: Ok got it, I think we should probably have some way of checking the beekeeper service health but I think that would be a different feature e.g. expose `systemctl status` output through odc-api

HariHive (approved):

mazar-farran:
Agreed to go forward with this on slack:

> I think the question in https://github.com/Hivemapper/odc-api/pull/342 is if this should report whether beekeeper is configured as enabled or if it's actually running (i.e. systemctl status beekeeper-plugin shows the service is running).
I think "configured as enabled" but not running would be a bug. We shouldn't report that Beekeeper is disabled in that scenario because the user would think it's a configuration issue

PR #354, opened 2025-03-14
Fixes for Uploading / Downloading / Packaging

punov:


src/services/lteFirmwareUpdateService.ts
```diff
@@ -394,6 +406,7 @@ export const LTEFirmwareUpdateService: IService = {
               event: 'DashcamLTEFirmwareUpdateEvent',
               message: 'Download failed',
             });
+            busy = false;
```
mazar-farran, line 409: This should not be necessary, the finally block is always hit and sets this to false

punov: Thanks, Mazar, you're right — updated

mazar-farran (approved):

PR #360, opened 2025-03-22
BK-454 Use fleetId in gnssAuth instead of fleet hash

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/BK-454/device-fleet-id-auth-with-hm

Note: This requires https://github.com/Hivemapper/beekeeper-plugin-service/pull/57. We will not merge this until the beekeeper-plugin pr is merged


src/services/gnssAuth.ts
```diff
@@ -184,16 +184,16 @@ export async function getGnssAuthCredentials() {
   };
 }
 
-export const getFleetHash = async () => {
-  const result = await fetch(`http://127.0.0.1:8000/settings/fleetHash`, {
-    method: 'POST',
+export const getFleetId = async () => {
+  const result = await fetch(`http://127.0.0.1:8000/info`, {
```
jimmdd, line 188: Does the info route return the fleet id?

mazar-farran: Yes, /info is defined here https://github.com/Hivemapper/beekeeper-plugin-service/pull/57

src/services/gnssAuth.ts
```diff
@@ -226,13 +226,13 @@ export const GnssAuthService: IService = {
         }
 
         const gnssAuthCredentials = await getGnssAuthCredentials();
-        const fleetHash = await getFleetHash();
+        const fleetId = await getFleetId();
```
jimmdd, line 229: NIL probably bad name, since getFleetId returns object of fleet id and fleet name

src/services/gnssAuth.ts
```diff
@@ -226,13 +226,13 @@ export const GnssAuthService: IService = {
...
 
         if (!gnssAuthCredentials) {
           console.log('GNSS Auth - No gnss auth credentials found');
           return;
-        } else if (!fleetHash) {
-          console.log('GNSS Auth - No fleet hash found');
+        } else if (!fleetId) {
```
jimmdd, line 234: this check might not do what you want, since getFleetId will always return object but fleetId inside the object might be null

jimmdd (changes requested):

src/services/gnssAuth.ts
mazar-farran: Good catch, thanks

src/services/gnssAuth.ts
mazar-farran: Updated name

jimmdd (approved):

PR #363, opened 2025-03-27
BK-454 Check last fleetid and reauth if it changed

mazar-farran:
Context: https://hivemapper.slack.com/archives/C08HFGDBGTG/p1743102131554739


src/services/gnssAuth.ts
```diff
@@ -215,6 +215,23 @@ export const GnssAuthService: IService = {
           return;
         }
 
+        const fleetIdResponse = await getFleetId();
+        const lastFleetId = JSON.parse(await getLastFleetId() || '{}');
+        console.log(`GNSS Auth - FleetId response: ${fleetIdResponse?.fleetId}, last fleetId: ${lastFleetId?.fleetId}`);
+        if (
+          fleetIdResponse &&
+          fleetIdResponse.fleetId !== undefined &&
```
harroyoc, line 223: maybe more like : `fleetIdResponse.fleetId != undefined` 
to include the null response

harroyoc: or do
`fleetIdResponse.fleetId !== undefined && fleetIdResponse.fleetId !== null` , right?

siiiiilvio: I think the validation for undefined is superfluous. The server can only return:

string | null | error (fleetIdResponse will not be truthy)

siiiiilvio: The case for null needs to be handled.

If the fleetId is null, we do not need to attempt the authentication.

mazar-farran: Null case is now handled before attempting auth

harroyoc (approved):

mazar-farran:
Note: persisting fleetId is probably not necessary but I'll look into it later

PR #364, opened 2025-03-28
EDGE-592 Support delta updates

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-592/update-ota-firmware-updates-to-support-delta-updates

The procedure is 
1. Check for full update url
2. From full update url + odc-api version, construct the delta update url
3. If this url exists, attempt to download and install the delta update
4. If we attempt and fail a delta update 3 times, fall back to full update


punov (approved):

mazar-farran:
minFirmwareBee targeting still works, need to verify that app targeting also works

mazar-farran:
app targeting worked

mazar-farran:
Note: firmware with this change will not support delta updates yet.

The firmware will also need to be updated to be a version 3 mender artifact. Once this is done, delta updates will be fully supported.

However, only dashcams with the updated mender client (5.4+) support version 3 artifacts. Other devices will fail to update

PR #372, opened 2025-04-16
Bugfix/remove check for plugins

HariHive:
This PR ensures that `systemctl` commands are always run irrespective of plugin state in DB


mazar-farran:
I think this is ok. Running `systemctl start` on an already started service looks like a no-op.

Is start/stop plugin always run on boot?

mazar-farran (approved):

punov (approved):

PR #375, opened 2025-04-22
Feature/support new route

HariHive:
This PR helps to get locations from GNSS concise table between start and end time for a given session


src/routes/gnssConcise.ts
```diff
@@ -16,15 +17,6 @@ router.get('/validGnssConcise/:n', async (req, res) => {
   }
 });
 
-router.get('/latestValid', async (req, res) => {
```
derekhive, line 19: We use this route in hive-cpp: https://github.com/Hivemapper/hive-cpp/blob/65b67df87e27bdd96aa66b19beea327894f4d77c/src/PositionApi.cpp#L12-L13

We can switch if needed, but if it's ok to leave, let's not delete

HariHive: Restored thank you!

derekhive (approved):

PR #381, opened 2025-05-10
EDGE-673 LTE Uploads: Handle 403 upload errors and remove old framekms

mazar-farran:
Context: https://hivemapper.slack.com/archives/C08RTB510SE/p1746836564684389
Ticket: https://linear.app/hivemapper/issue/EDGE-673/dashcam-metadata-should-not-return-framekms-older-than-10-days


harroyoc (approved):

HariHive:
@mazar-farran  If understand correctly does **CleanupFramekmService** service runs every 2 seconds ?

src/services/cleanupFramekmService.ts
```diff
@@ -0,0 +1,64 @@
...
+          await unlinkIfExists(fileName);
+          console.log(`Deleting file: ${framekmFileName} because it's older than 10 days`);
+          await unlinkIfExists(framekmFileName);
+        }
+      }
+    } catch (e: unknown) {
+      console.log(e);
+    }
+  },
+  interval: 2000,
```
HariHive, line 63: Feels like 2seconds is too frequent

src/services/getFrameKmsService.ts
```diff
@@ -458,9 +458,9 @@ async function processFrameKms(requestRejectionMessages: string[]) {
         console.log('FrameKmUploadError', e);
         // We're assuming all 400 errors are due to expired tokens, but that's not necessarily true
         // We should add an error message check in the future to be safe.
-        if (e.status === 400) {
+        if ((e.status === 400 || e.status === 403)) {
           console.log(
-            'Token possibly expired, resetting framekm entry to initial state',
+            `Token possibly expired with error ${e.status}, resetting framekm entry to initial state`,
```
HariHive, line 463: Did we confirm here that fkm will get a new upload URL for these cases?

mazar-farran: This will reset the framekm entry so it will repeat the initialization -> request -> upload steps

src/services/cleanupFramekmService.ts
mazar-farran: Updated to 60 seconds

HariHive:
@mazar-farran  please correct me if wrong, the clean up of fkms service feels like once per boot kind of operation, do we need a periodic service for it?
can we add it to a place where this function runs only once for everytime GPs lock is acquired ?

src/services/getFrameKmsService.ts
HariHive: Hmm, but line 466 deletes the Fkm right ?

HariHive:
Not super familier with short fkms but should we also clean up from : `export const FRAMEKM_SHORT_ROOT_FOLDER = '/mnt/data/framekm_short';
`

src/services/getFrameKmsService.ts
mazar-farran: It deletes the entry, not the framekm. The framekm entry will be reinitialized and we will proceed to the request stage. The framekm entry and data will be deleted if it's rejected in the request stage

mazar-farran:
> Not super familier with short fkms but should we also clean up from : `export const FRAMEKM_SHORT_ROOT_FOLDER = '/mnt/data/framekm_short'; `

Good catch, thanks, updated the function

mazar-farran:
> @mazar-farran please correct me if wrong, the clean up of fkms service feels like once per boot kind of operation, do we need a periodic service for it? can we add it to a place where this function runs only once for everytime GPs lock is acquired ?

This would be better but I don't think we have something like that in odc-api, it looks like all gps checks are periodic.  It might be worth adding hooks for events like this but, I think internally we'd do the same thing (poll for gps and call all callbacks if we get a lock)

I added a flag to only run the code in the service once

HariHive (approved):

HariHive:
> > @mazar-farran please correct me if wrong, the clean up of fkms service feels like once per boot kind of operation, do we need a periodic service for it? can we add it to a place where this function runs only once for everytime GPs lock is acquired ?
> 
> This would be better but I don't think we have something like that in odc-api, it looks like all gps checks are periodic. It might be worth adding hooks for events like this but, I think internally we'd do the same thing (poll for gps and call all callbacks if we get a lock)
> 
> I added a flag to only run the code in the service once

thank you

PR #382, opened 2025-05-13
EDGE-673 Add instrumentation to cleanup service, simplify logic

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-673/dashcam-metadata-should-not-return-framekms-older-than-10-days


src/services/cleanupFramekmService.ts
```diff
@@ -53,21 +54,34 @@ export const CleanupFramekmService: IService = {
...
       
         const timestampMs = new Date(formatted).getTime();
-        console.log("Timestamp in ms:", timestampMs);
-        dates.push({fileName: metadatafile, timestampMs: timestampMs});
-      }
-      
-      for (const {fileName, timestampMs} of dates) {
-        const basename = path.basename(fileName, '.json');
+        if (isNaN(timestampMs)) {
+          console.log(`Invalid date format in file name: ${metadatafile}`);
```
HariHive, line 58: if it's invalid should we still delete ?

src/services/cleanupFramekmService.ts
```diff
@@ -53,21 +54,34 @@ export const CleanupFramekmService: IService = {
...
+          continue;
+        }
+
         if (currentGpsTime - timestampMs > 10 * 24 * 60 * 60 * 1000) {
```
HariHive, line 62: sorry for flagging it late, can we move this value 10 * 24 * 60 * 60 * 1000 to a const ?

src/services/cleanupFramekmService.ts
```diff
@@ -53,21 +54,34 @@ export const CleanupFramekmService: IService = {
...
-          console.log(`Deleting framekm ${basename} because it's older than 10 days`);
-          await cleanupFrameKm(fileName);
+          deletedFrameKms.push(metadatafile);
+          console.log(`Deleting framekm ${metadatafile} because it's older than 10 days`);
+          await cleanupFrameKm(metadatafile);
         }
       }
 
+      Instrumentation.add({
+        event: 'DashcamFrameKmCleanup',
```
HariHive, line 70: Should we call them : DashcamFrameKmCleanupSuccess and DashcamFrameKmCleanupFailure?

HariHive (changes requested):

src/services/cleanupFramekmService.ts
mazar-farran: added

src/services/cleanupFramekmService.ts
mazar-farran: done

src/services/cleanupFramekmService.ts
mazar-farran: done

HariHive (approved):

mazar-farran:
Hi @HariHive I changed this to skip framekms which it failed to parse.  I think it's better to err on the side of caution because we don't want to delete a framekm we weren't meant to delete

It will report any framekms it skipped in the instrumentation

PR #383, opened 2025-05-13
EDGE-722 Add device lock

mazar-farran:
POST /setPluginLockState
true = lock
false = unlock

If locked, /enablePlugin and /disablePlugin return a 403

This does NOT prevent someone motivated enough from disabling beekeeper, it only adds an extra hidden switch for enabling/disabling these apis.


MattPeytonHivemapper (commented):
Looks fine to me. However, another firmware team person should double check. e.g. @HariHive

HariHive:
@mazar-farran  if I am not wrong this locks all plugins and not only beekeeper right ?
I know right now we only have BK plugin 
Are we fine with this ?

HariHive:
I think we only want to lock bk plugin for now

mazar-farran:
> I think we only want to lock bk plugin for now

This would be a little more complex, we would need to store an entry per-plugin to lock it instead of a global config.  Is there a requirement for this (locking BK but not wanting to lock other plugins)?

HariHive:
Discussed offline with Mazar, we will change Plugins table on ODC API to add a new column which would tell if the plugin should be locked or not

mazar-farran:
Evan seems ok with the global plugin approach so we will stick with this for now because it's simpler. If we need to extend this to support per-plugin locking we can add an additional check later

MattPeytonHivemapper:
> Evan seems ok with the global plugin approach so we will stick with this for now because it's simpler. If we need to extend this to support per-plugin locking we can add an additional check later

Alright, given this, ill leave it as-is

HariHive (approved):

MattPeytonHivemapper:
So sorry I merged this, thought I was looking at a different PR!

PR #384, opened 2025-05-13
BK-604: Add new Vision AIEvents endpoint

dhruvk-hm:
EDIT: Cancelling this PR. Ignore this.

1. Add median speed calculations for frames. Get the median frames form all gps entries between 2 frames. 1 entry every 0.125 seconds.
2. Agg endpoint to query vision events for beekeeper.


src/sqlite/framekm.ts
```diff
@@ -2,7 +2,11 @@ import { getAsync, runAsync } from './index';
 import { FrameKM, FrameKmRecord } from 'types/sqlite';
```
dhruvk-hm, line 2: Most of this is autolint. The only real change here is `median_speed`

src/sqlite/aiEvents.ts
```diff
@@ -1,14 +1,37 @@
...
+      },
+    );
+  });
+};
+
+export const fetchVisionAiEventsSince = async (
+  timestamp: string,
+): Promise<VisionAIEventRecord[]> => {
+  // TODO: Remove limit once debug events removed.
+  const query = `SELECT * FROM ai_events_vision WHERE ts >= ? ORDER BY ts DESC LIMIT 5`;
```
HariHive, line 30: Is it intended to limit to 5?

src/sqlite/index.ts
```diff
@@ -280,6 +286,8 @@ export const performSoftMigrations = async (db?: Database): Promise<void> => {
   const migrationCommands = [
     `ALTER TABLE framekms ADD COLUMN orientation INTEGER DEFAULT 1;`,
     `ALTER TABLE packed_framekms ADD COLUMN orientation INTEGER DEFAULT 1;`,
+    `ALTER TABLE framekms ADD COLUMN median_speed REAL DEFAULT 0;`,
```
HariHive, line 289: Lets test that fkms uploads on app are not broken

dhruvk-hm: Cancelling this. Thanks for bringing up these concerns.

PR #404, opened 2025-05-28
Feature/wifi only bee

HariHive:


punov (approved):

src/routes/framekm.ts
```diff
@@ -201,12 +202,7 @@ router.get('/total', async (req: Request, res: Response) => {
 router.delete('/:name', async (req: Request, res: Response) => {
   try {
     const name = req.params.name;
-    if (existsSync(FRAMEKM_ROOT_FOLDER + '/' + name)) {
-      rmSync(FRAMEKM_ROOT_FOLDER + '/' + name);
-    }
-    if (existsSync(FRAMEKM_SHORT_ROOT_FOLDER + '/' + name)) {
-      rmSync(FRAMEKM_SHORT_ROOT_FOLDER + '/' + name);
-    }
+    await cleanup(name);
```
mazar-farran, line 205: Is this tested? I'm not sure what would happen if a framekm entry was removed in the middle of uploading LTE frames

mazar-farran: sorry thought I made this comment this morning

src/config/bee.ts
```diff
@@ -99,8 +99,8 @@ export const CMD = {
   READ_DEVICE_INFO:
     '/opt/dashcam/bin/eeprom_access.py -r -f /tmp/dump.bin -o 0 -l 30',
   DEVICE_SSID: 'cat /sys/firmware/devicetree/base/serial-number',
-  DELETE_FIRMWARE_FILES:
-    `rm -rf ${LTE_FIRMWARE_ROOT}*.raucb && rm -rf ${APP_FIRMWARE_ROOT}*.raucb`,
+  DELETE_FIRMWARE_FILES: `rm -rf ${LTE_FIRMWARE_ROOT}*.raucb && rm -rf ${APP_FIRMWARE_ROOT}*.raucb`,
```
evanmoss, line 102: why are there different firmwares?

HariHive: cc @mazar-farran  to be give more context on this I believe this is a change from lint

mazar-farran: Yes this looks like a formatting change

mazar-farran:
I think this PR looks ok

Some notes:

There is a `cleanup()` bugfix introduced by @punov that is not really related to this feature request.  Discussed offline and we think it should be low risk

This PR does NOT change `getPluginState` to report `false` if the device is Wi-Fi only. Discussed with @HariHive and @Masaya-RT . `getPluginState` is a source of truth on whether or not a plugin is enabled. It should never report "false" if beekeeper plugin is actually enabled.

Instead, we already provide a `isWiFiDevice` flag in odc-api `/info`. The app or test should check this + `getPluginState` and make a decision to display beekeeper UI from that

cc @evanmoss

PR #408, opened 2025-05-29
EDGE-774 Add import sorter to pre-commit

mazar-farran:


mazar-farran:
I'm not very happy with the current state of this, it sorts imports but doesn't remove unnecessary ones, which is what I was mostly interested in.

It looks like only eslint supports removing imports via a plugin, but I was not able to get this to work.

PR #413, opened 2025-06-02
Fix FrameKM ID generation

punov:
FrameKM was generated as upticking numeric number,
Now it is fixed to be based on unix timestamp of the first record of FrameKM


derekhive (approved):
row.time comes from GNSS records where time is a UTC time based on GNSS from the NAV_PVT message. Should be non-conflicting as soon as there is a GNSS fix

PR #418, opened 2025-06-04
EDGE-780 Change plugin to use Etag (md5) instead of metadata (sha256)

mazar-farran:


punov (approved):

mazar-farran:
Does not actually work with multipart uploads

> If an object is created by either the Multipart Upload or Part Copy operation, the ETag is not an MD5 digest, regardless of the method of encryption. If an object is larger than 16 MB, the AWS Management Console will upload or copy that object as a Multipart Upload, and therefore the ETag will not be an MD5 digest.

PR #422, opened 2025-06-05
EDGE-780 Change plugin check use new /plugins url

mazar-farran:
Uses the plugins url implemented in https://github.com/Hivemapper/network/pull/4288


punov (approved):
Looks great to me,
@mazar-farran do you mind adding an instrumentation event like DashcamUpdatedPlugin or similar, so we can also track it?

PR #428, opened 2025-06-11
make service call async

HariHive:
This PR has 
1. /zoo/refresh call async, so that app does not wait for long time till it receives response
2. App hash version numbers to the /zoo/models call so that app can use version numbers to compare with server


HariHive:
image
Tested on postman

harroyoc (approved):

punov (approved):

PR #447, opened 2025-07-01
EDGE-866 Add service for non-beekeeper GNSS Auth

mazar-farran:
Registers the current user with the current device.  If the current cookie stops working, we will retrieve the credentials using gnss auth.

This addition is in appGnssAuth.ts. 

gnssAuth.ts (now called bkGnssAuth.ts) was rewritten to use more common code between these two (most importantly we removed callAPI and replaced it with callHmApi). This requires extensive testing but I think it was necessary to avoid making the code base too difficult to work with.

User story:
1. When I connect my wallet to the app, and my phone to the dashcam, the dashcam will get configured with my credentials.
3. The dashcam will be able to map using LTE indefinitely after this. If the current cookie expires, the dashcam can get a new one without the app.
4. If I change wallets in the app and connect to the dashcam again, my dashcam will use the new wallet.

Related: https://github.com/Hivemapper/network/pull/4444


src/util/api.ts
```diff
@@ -52,7 +52,7 @@ export async function callHmApi(
 
       const newOptions: RequestInit = {
         ...options,
-        agent: getAgent(),
```
mazar-farran, line 55: TODO: remove

punov (approved):

mazar-farran:
Hi @HariHive could you review the changes to `gnssAuth.ts` ? I will try to test this path manually

mazar-farran:
I'm realizing that there's a potential race condition here that is very difficult to address

1. Device is on BK and has BK auth
2. BK is disabled on device in the middle of authenticating
3. Cookie is cleared
4. In-progress authentication succeeds and sets the cookie
5. Now the device has a BK cookie while in non-bk mode

This would in theory happen with the latest `bee`

Update: added additional guards in the backend to prevent registering a bk user as an app lte user

src/services/bkGnssAuth.ts
```diff
@@ -0,0 +1,216 @@
...
+  if (fetchJson?.loginProvider !== 'gnss') {
+    console.log(
+      'BK GNSS Auth - Invalid login provider:',
+      fetchJson?.loginProvider,
+    );
+    return false;
+  }
+
+  console.log('Authenticated with Hivemapper');
+  console.log('Username is', fetchJson?.username || 'unknown');
```
HariHive, line 75: Can we combine above two logs into one ?

src/util/gnssAuth.ts
```diff
@@ -0,0 +1,87 @@
...
+  passUserInfoToStorage,
+} from 'services/getFrameKmsService';
+import { Response } from 'node-fetch';
+
+export async function getGnssAuthCredentials() {
+  let gnssAuth = [];
+
+  try {
+    gnssAuth = (await getAsync(
+      'SELECT buffer, buffer_hash, gnss_session_id, signature FROM gnss_auth ORDER BY id DESC LIMIT 1',
```
HariHive, line 18: Can we add this to possibly a file under folder sqlite ?

mazar-farran: done

src/services/appGnssAuth.ts
```diff
@@ -0,0 +1,148 @@
...
+let busy = false;
+export const LteGnssAuthService: IService = {
+  execute: async () => {
+    if (busy) {
+      return;
+    }
+    try {
+      busy = true;
+
+      if (!(await isLTEHealthy())) {
```
HariHive, line 55: if it is a wifi Device should we return here ?

src/services/bkGnssAuth.ts
```diff
@@ -0,0 +1,216 @@
...
+import { callHmApi } from 'util/api';
+import { isLTEHealthy } from 'util/lte';
+
+let forceGnssAuth = true; // Should always get a new cookie on boot if beekeeper enabled
+export const forceGnssLogin = async () => {
+  forceGnssAuth = true;
+};
+
+export const getFleetId = async () => {
+  const result = await fetch(`http://127.0.0.1:8000/info`, {
```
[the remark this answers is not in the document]
mazar-farran: Moved to constant

mazar-farran:
Tested with multiple CTP users https://hivemapper.slack.com/archives/C07BW6NC7C6/p1752534250567649

BK user was confirmed uploading
3 LTE auth users were confirmed uploaded
1 user with no gnss public key was confirmed uploading
1 user with wi-fi only was on vacation, could not confirm

With these results I think this is good to pass to RC

HariHive (approved):

mazar-farran:
For future reference: you can verify which users are registered in the backend with  `> await LteGnssAuth.find()`

PR #471, opened 2025-08-12
EDGE-977 EDGE-978 Combined wifi client upload mode config

mazar-farran:
This will hopefully be a centralized way to know if the device is in LTE, app, or wifi mode.

Changes:

Added config apis for upload mode.

If no config present, device will set config to LTE on LTE devices, and APP on wifi devices

Removed all fallback code in getFrameKMService. Device will always attempt to upload if not configured in app mode.

Internet LED is now on if

Mode is not APP
Able to ping hivemapper
Has been less than X seconds since last successful ping
/info now reports the LED on/off status instead of the internal gps/lte state. This way app is always in sync with LED

Removed non-bee code from heartbeat.ts


src/index.ts
```diff
@@ -1,14 +1,9 @@
 import express, { Application } from 'express';
 import router from './routes';
-import { Server } from 'http';
+import { Server as HttpServer } from 'http';
+import { Server as HttpsServer } from 'https';
```
mazar-farran, line 4: Please remove https server.  It is not used

src/types/wifiClient.ts
```diff
@@ -26,3 +26,5 @@ export interface NetworkScanInfo {
   security: WifiClientSecurity;
   freq: number;
```
mazar-farran, line 27: Do we still need freq?

src/util/framekm.ts
```diff
@@ -29,18 +29,18 @@ import {
   DetectionsByFrame,
```
mazar-farran, line 29: This file seems to have a lot of unrelated changes, possibly needs to be synced to bee?

src/util/lte.ts
```diff
@@ -19,7 +22,7 @@ export async function isLTEServiceEnabled() {
 }
 
 export async function isInternetHealthy() {
-  return (await isLTEHealthy()) || isWifiClientHealthy();
+  return (await isLTEHealthy()) || (await isWifiClientHealthy());
```
mazar-farran, line 25: isWifiClientHealhty is not an async function

HariHive: good catch fixed

src/util/lte.ts
```diff
@@ -144,9 +146,20 @@ export const getLTEServiceStatus = async () => {
...
+  let last_operator_used: string | null = null;
+  try {
+    last_operator_used = await fs.readFile(
+      '/data/recording/carrier_info.txt',
+      'utf-8',
+    );
+  } catch (e) {
+    console.log('Error reading last operator used:', e);
+    last_operator_used = null;
+  }
```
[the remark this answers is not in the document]
HariHive: removed

mazar-farran:
LGTM, I can't approve my own PR though

HariHive (approved):

PR #474, opened 2025-08-14
EDGE-1014 Delete duplicate framekms

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-1014/fix-duplicate-framekms-not-getting-deleted

Delete the framekm if it fails to commit ~unless it's unauthorized~

https://github.com/Hivemapper/network/blob/a97e5a1d8239d18dde7149c7c291a329f1cbb257/server/src/routes/framekm.ts#L124


mazar-farran:
@punov the issue with making an exception for unauthorized is that the user will not understand why we are not uploading that framekm, it will stay stuck. I would prefer to not make an exception for this given that this error seems to be rare to  nonexistent, and I don't want to have callouts for expected behavior

punov (approved):

mazar-farran:
Tested locally and the duplicate framekm was deleted

PR #450, opened 2025-07-07
EDGE-878 Wifi client mode

mazar-farran:
`GET /settings`
Purpose: Retrieves the current Wi-Fi client settings.

Fetches settings (ssid, password, enabled) from persistent storage.

`POST /settings`
Purpose: Updates the Wi-Fi client settings.

Input: ssid, password, and enabled ('true' or 'false') in the request body.

Saves the new settings to persistent storage.

`GET /status`
Purpose: Returns the current connection status of the Wi-Fi client.

`GET /scan`
Purpose: Lists available Wi-Fi networks.

Returns an array of detected networks.

`GET /reset`
Purpose: Resets the Wi-Fi client with stored configuration.

Calls resetWifiClient() with the saved ssid and password.

- [x] Add a link to the ticket in the PR title or add a description of work in the PR title
- [x] dashcam firmware was loaded on a device and successfully runs

PR #477, opened 2025-08-15
EDGE-878 Add initial wifi client functionality

mazar-farran:
Initial version of wifi client.  Merging to reduce delta with the test branch.  This commit should not change any bee behavior as long as the new endpoints are not used

Original PR: https://github.com/Hivemapper/odc-api/pull/450


mazar-farran:
Test drive looked good https://hivemapper.slack.com/archives/C07BW6NC7C6/p1755290319980769

mazar-farran:
Merging now because this should be a no-impact change, and drive testing was good

PR #484, opened 2025-09-02
fetch default trip trim distance from server

HariHive:
- pulls trip trimming from the server
- Added trip trimming part of motion controller back in
    - loops until drive session is ready (time received from GNSS)
    - starts drive session (only thing in start function is doing trip trimming)
    - stop iterating
    
 - Check that's preventing multiple frames from being trimmed if rebooting is whether it's been 4 minutes since the last trim happened (existing logic)
 - That logic didn't actually work since the last trimmed time was always being set to zero on boot (so the time that was supposed to carry over from the previous drive session was never actually being carried over)
 - Changed so that DX = 8.0 is always being used (much easier than reading from table imo and I think we're ok with a 50% trim loss)


src/services/initialiseConfig.ts
```diff
@@ -46,6 +60,19 @@ export const InitialiseConfigService: IService = {
         );
       }
     }
+    if (!wereServerDefaultsFetchedAtleastOnce) {
+      try {
+        //Trip trim distance defaults from server
+        const tripTrimDistance = await getTripTrimServerConfig();
```
derekhive, line 66: If this function fails do you want to handle it differently? e.g. not set and don't mark as fetched once?

HariHive: if response is not okay we throw it, so `wereServerDefaultsFetchedAtleastOnce` will be true only if it is fetched atleast once

src/services/initialiseConfig.ts
```diff
@@ -1,24 +1,38 @@
...
       const defaultConfig = getDefaultConfig();
       defaultConfig.isInitialised = true;
+      const defaultServerTrimDistance = await getConfig(
+        DEFAULT_TRIM_DISTANCE_FROM_SERVER_KEY,
+      );
+      //if we ever set the trip trim distance from server, use that as default
+      if (defaultServerTrimDistance) {
+        defaultConfig.TrimDistance = defaultServerTrimDistance;
+      } else {
+        defaultConfig.TrimDistance = DEFAULT_TRIP_TRIM_DISTANCE;
```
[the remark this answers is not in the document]
HariHive: `defaultServerTrimDistance` => this will never be set for wifi only devices so as a safe fallback I have added else condition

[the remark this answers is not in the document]
punov: I agree with @derekhive , 
```const defaultConfig = getDefaultConfig()```
should return the default DEFAULT_TRIP_TRIM_DISTANCE value,
if it doesn't — there's a bug somewhere!

File: src/util/tripTrim.ts [hunk 3]
```diff
@@ -0,0 +1,14 @@
+import { callHmApi } from './api';
+
+export const getTripTrimServerConfig = async () => {
+  const response = await callHmApi('/companion/globalconfig', {
```
punov, line 4: Can it be done as part of `/companion/dashcamconfig/v2` ?
We already have a service `getConfigService` that pulls it,
it will make sense to me

src/services/initialiseConfig.ts
HariHive: updated to set to true only when valid number is fetched from server

src/services/initialiseConfig.ts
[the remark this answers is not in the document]
HariHive: updated

derekhive (approved):
validated as v5.7.99 firmware

PR #486, opened 2025-09-04
Stop log request retries and reduce console logs

dhruvk-hm:
Log request retries create too many logs, especially on Wi-Fi devices.


derekhive (approved):
validated in 5.7.94 firmware

PR #497, opened 2025-09-22
Add `/metadata/all` endpoint

HariHive:
`/metadata` preserves the old behavior of hiding metadata when the device is in LTE mode.

`/metadata/all` shows all metadata unfiltered.

The app will use `/metadata/all` from now on, `/metadata` is kept for legacy reasons

Thank you @mazar-farran


mazar-farran (dismissed):
lgtm

mazar-farran (approved):

PR #535, opened 2025-10-28
Derek/latest position endpoint

derekhive:
Adds latest position endpoint


src/routes/index.ts
```diff
@@ -35,6 +35,7 @@ import previewRouter from './preview';
 import pluginRouter from './plugin';
 import aiEventsRouter from './aiEvents';
 import landmarksRouter from './landmarks';
+import PositionRouter from './position';
```
mazar-farran: could this be `positionRouter` ?

src/sqlite/gnssConcise.ts
```diff
@@ -84,7 +84,7 @@ export const fetchLatestValidGnssConciseRecordForSession = async (): Promise<
 
   // Now, retrieve the latest record from gnss where the session matches the one from gnss
   const query = `
-    SELECT id, utc_time, latitude, longitude, altitude, speed AS estimated_speed, hdop, gdop, eph
+    SELECT id, utc_time, latitude, longitude, altitude, ROUND(speed,3) AS estimated_speed, hdop, gdop, eph
```
mazar-farran: What is the reason for this change? Could this affect other things that use `fetchLatestValidGnssConciseRecordForSession` ?

derekhive: It's only used for the here-plugin. Previously it was giving maybe 14 decimal places for speed and that's misrepresenting how precisely we estimate speed. We don't know it down to the nanosecond

mazar-farran: Derek confirmed only the user plugin uses this, not likely this will break our firmware

mazar-farran (approved):

PR #545, opened 2025-11-06
EDGE-1212 add plugin stats telemetry

mazar-farran:
Overview:
   1. If an endpoint is marked with 'trackForLogging', we will update a counter for it every time it is hit.
   2. After 5 minutes, the counter will be written to the DB
   3. After N hours, we will send the aggregated counts of the last week, day by day, to the backend.
   4. If we have entries in the DB that are older than 1 week, we purge it.  
     - Open question: should we purge as soon as we upload, or keep this data around for a week?
   5. Backend is not implemented yet, but it will store these 24 hour aggregates as mongo documents, replacing them if it gets a duplicate (this way, the current day gets updated hourly).

Limitations:
Even though we are recording the plugin name, this currently has no way to tell which plugin a request is coming from if any. For now it is just hard coded

Possible alternatives:
Do we need to aggregate on the device? Could we just send the raw data entries and let the backend worry about aggregating them?
Do we need to hold on to old entries once they've been sent to the backend?

Requires: https://github.com/Hivemapper/network/pull/5037


mazar-farran:
Some extra guarding is probably needed on the backend side. E.g. we expect the updates to be monotonic, so if we get an update to the count that is smaller than the current value, should probably discard (I'm concerned we might accidentally update some entries to 0 in some edge cases)

mazar-farran:
~Note: the new sqlite table is not indexed. We could add that in a follow up if needed~ Added index on start_ts

mazar-farran:
Stalled waiting for requirements changes

derekhive (approved):

PR #554, opened 2025-11-13
Filter sql on confidence value

derekhive:


punov (approved):

mazar-farran:
No longer needed, #555 is the latest version of this

PR #568, opened 2025-11-24
Sort latest lane by descending

derekhive:
- The lane counts endpoint should give the newest value rather than the oldest


yqin100 (approved):
LGTM

PR #573, opened 2025-12-05
EDGE-1303 Change /upload endpoint to take a landmark ID instead of a path

mazar-farran:


mazar-farran:
I've verified locally that this works

Let's merge this to `bee` without a CTP cycle, this API is not currently used so it will not impact anyone

derekhive (approved):

src/routes/landmarks.ts
```diff
@@ -642,9 +642,11 @@ router.put('/upload/', async (req, res) => {
   }
 
   try {
-    if (!existsSync(filePath)) {
-      return res.status(404).send({ error: `File not found: ${filePath}` });
+    const paths = await getObservationsImagePathsByLandmarkId(landmarkId);
+    if (!paths.length) {
+      return res.status(404).send({ error: `No valid images found to upload` });
     }
+    const filePath = paths[0];
```
derekhive: paths[paths.length - 1] to get the most recent (largest pixel area) image

mazar-farran: done

PR #578, opened 2025-12-10
EDGE-1307 Add unit tests for landmarks

mazar-farran:
TODO: test with subclass filters


mazar-farran:
@punov  @derekhive I wrote the above to get the ball rolling but I hope in the future we can start writing these types of tests on important parts of the code base.

To make unit testing easy, we need to write code that separates I/O from business logic. Acquiring the db, files, etc. are ideally done in the main function or in the endpoint handles. 

That way we could pass in mock dbs, files, etc., instead of mocking internal functions like in https://github.com/Hivemapper/odc-api/pull/558/files#diff-030ae8d84fa96d35fd2325dff2cec4ecdf78de1236476d24292ee5711e246307. The linked approach is fragile and needs more understanding of the implementation.

derekhive (approved):

punov (approved):

PR #584, opened 2025-12-12
EDGE-1311 Add gateway apis

mazar-farran:
This introduces a `gatewayFetch` api, which is an alternative to `fetch`. It caches http requests in a database to be fulfilled by an external forwarder (i.e. the Bee app).  It is only used when the device is in "app" mode.

Right now this is only used as part of the /forward api. Which user plugins use to send their requests.

Related to: https://github.com/Hivemapper/hivemapper-hdc-os/pull/637


mazar-farran:
Tested in CTP with 6.66.1. It did not cause any regressions

mazar-farran:
AI Generated overview (lgtm)

# Gateway Service Summary

This document summarizes the changes introduced by the `origin/add-gateway-service` branch, relative to `origin/bee`.

## Overview
The Gateway Service introduces a mechanism to cache and fulfill HTTP requests through an external forwarder (e.g., the Bee app). This is primarily used when the device is in **"app" mode**, enabling connectivity through a mobile device rather than direct LTE.

## Key Components

### 1. `gatewayFetch` API
Introduced in `src/util/gateway.ts`, this is an alternative to the standard `fetch`.
- **Behavior**:
    - **App Mode**: Instead of sending a network request directly, it enqueues the request in a local database and polls for a response from the external forwarder.
    - **Standard Mode**: Falls back to a normal `fetch` with a `Bee-Gateway-Sender: bee` header.

### 2. Database Caching
Two new SQLite-backed entities are used to manage the request/response lifecycle:
- **`GatewayRequests`**: Stores enqueued requests, including method, URL, headers, and base64-encoded body.
- **`GatewayResponses`**: Stores the responses received from the forwarder.
- **Cleanup**: Both requests and responses are automatically cleaned up if they exceed a 30-second TTL.

### 3. API Endpoints (`/api/1/gateway`)
The service exposes several routes in `src/routes/gateway.ts`:
- `POST /submit`: Enqueues a request and returns a `jobId`.
- `GET /pending`: Returns a list of pending requests for the external forwarder to process.
- `POST /proxyResponse`: Receives the response from the forwarder and saves it to the database.
- `GET /status/:jobId`: Polls for the result of a specific job.
- `POST /forward`: A high-level endpoint that wraps `gatewayFetch` for external requests.

### 4. Integration
Right now, this service is primarily utilized by the `/forward` API, which user plugins will use to communicate with external services (introduced by https://github.com/Hivemapper/hivemapper-hdc-os/pull/637). This hopefully will allow plugins to function regardless of whether the device has a direct internet connection or is tethered to an app.

harroyoc (approved):

PR #588, opened 2025-12-19
EDGE-1315 More generic plugin

mazar-farran:
Related to: https://github.com/Hivemapper/network/pull/5117

Summary of changes:
1. Instead of updating a hardcoded plugin, we have a for loop over all enabled plugins
2. Instead of checking isUserPluginUser (boolean) to determine if we should enable the plugin, we check every plugin's PluginConfig to decide if we should enable/update it.
3. PluginConfig now comes from the user config instead of a hardcoded config file.


src/services/getUserConfigService.ts
```diff
@@ -95,5 +95,5 @@ export const GetUserConfigService: IService = {
       console.log('GetUserConfigService error', e);
     }
   },
-  interval: 600000, // re-check every 10 minutes
+  interval: 5000, // re-check every 10 minutes
```
punov: just not in sync with the comment anymore

src/util/lte.ts
```diff
@@ -10,7 +10,7 @@ import { getConfig } from 'sqlite/config';
 import { UPLOAD_MODE_KEY } from 'types/motionModel';
 import path from 'path';
 
-export const BASE_URL = 'https://hivemapper.com';
+export const BASE_URL = 'http://192.168.197.54:10000';
```
punov: @mazar-farran making sure this one is expected — probably a leftover from local testing?

punov (dismissed):

src/services/pluginUpdateService.ts
```diff
@@ -71,222 +71,217 @@ export const isUpdatablePluginEnabled = async (pluginName: string) => {
...
+export const PluginUpdateService: IService = {
+  execute: async () => {
+    if (busy) {
+      return;
+    }
+    busy = true;
+
+    try {
+      for (const plugin of updatablePlugins) {
+        await updatePlugin(plugin);
```
HariHive: Does it make sense to add try catch to `await updatePlugin(plugin);`
so in case one of the plugins failed to update for some reason, for loop still proceeds to update next plugin ?

HariHive (approved):

mazar-farran:
This is a low risk change so we are merging now

PR #591, opened 2025-12-31
routes to enable and disable depth

HariHive:


src/routes/cache.ts
```diff
@@ -84,4 +87,26 @@ router.get('/status', async (req: Request, res: Response) => {
   }
 });
 
+router.post('/enableDepthFlag', async (req: Request, res: Response) => {
+  try {
+    await setConfig(DEPTH_FLAG_KEY, true);
```
derekhive: Add `execSync('systemctl restart map-ai');` after this line so map ai restarts

src/routes/cache.ts
```diff
@@ -84,4 +87,26 @@ router.get('/status', async (req: Request, res: Response) => {
...
+      success: true
+    });
+  } catch (error: unknown) {
+    res.status(500).json({ error });
+  }
+});
+
+router.post('/disableDepthFlag', async (req: Request, res: Response) => {
+  try {
+    await setConfig(DEPTH_FLAG_KEY, false);
```
derekhive: Add `execSync('systemctl restart map-ai');` after this line so map ai restarts

src/routes/cache.ts
HariHive: done

derekhive (approved):

PR #593, opened 2026-01-03
EDGE-1308 Add typeORM, replace config queries with typeORM queries.

mazar-farran:
Ticket: https://linear.app/hivemapper/issue/EDGE-1308/investigate-using-orms-for-sqlite-in-odc-api

Summary of changes:
1. Add typeORM and make necessary changes to package to allow this.
2. Integrate typeORM with pre-existing sqlite3 setup
    - typeORM is fully responsible now for creating the tables and migrations, but sqlite3 still does some initialization (e.g. plugin init)
       - (exception: some tables like landmarks are created by map-ai, we are not touching this for now)
    - most endpoints/services still use sqlite3 for reading/updating.
3. Added new interface for Config which uses typeORM (`getConfigSingle`, `getConfigMulti`, `setConfigDataSource`), which are used by the `/config` endpoints.  
4. Old apis (e.g. `getAsync`, `runAsync`) have been deprecated`

We should create all new tables using typeORM entities from now on.  

Queries should be written with `repo.find()/update/etc.` if possible, or with `repo.queryBuilder()` for more complex queries. As a last resort, we can use `repo.runQuery()` to run sqlite directly.

New sql functions should have unit test coverage. Gen AI is good for creating these tests but make sure the tests make sense and properly catch what they say they are validating.

Generating migrations is kind of clunky.  I used this file to generate the migrations, but sometimes the migration generated is overkill (e.g. copy contents to a new table instead of using ALTER TABLE).

[ormconfig.ts](https://github.com/user-attachments/files/24457863/ormconfig.ts)


derekhive (approved):

HariHive (approved):

mazar-farran:
Tested and verified in 6.57.1

PR #597, opened 2026-01-09
EDGE-1320 Add new directory parameter for plugins

mazar-farran:
This lets multiple plugins share a path, which lets us re-use the service file (i.e. we don't have to make new recipes for every customer).

Context: https://hivemapper.slack.com/archives/C08U67BJ9Q9/p1767913994754019

Related to: https://github.com/Hivemapper/network/pull/5125


derekhive (approved):

mazar-farran:
Tested locally and seems to work

PR #605, opened 2026-01-28
EDGE-1311 Use callHmApiWithGateway for landmarkProxyService

mazar-farran:
Allows us to send landmark audits with the app only


mazar-farran:
I have validated that this works locally

Procedure:
1. I examined landmarkProxyService.ts to get a sense of what it was doing. Luckily this is a very simple service that sends a PUT request to the backend every 5 minutes. Unfortunately, it was more difficult to determine the shape of the payload than it should have been due to the use of the 'any' type.
2. After examining the endpoint on the network side https://github.com/Hivemapper/network/blob/main/server/src/routes/plugins.ts#L191, I found a function `validateAuditRecordPayload` that strictly defined the shape of the payload. I asked chatgpt to create a dummy payload from this snippet to save time.
3. I modified landmarkProxyService to send a hardcoded landmark to avoid having to get real data, and reduce the interval to 5 seconds. (https://github.com/Hivemapper/odc-api/pull/605/changes/0a557df8cb49bbcc4443a49fe20f3fddf7f09b67). I also modified the service to print the session ID. 
4. I found the landmark on the backend using the repl `x = await PluginAuditRecord.find({session:'<session-string>'});`
5. After confirming this worked, I replaced `callHMApi` with `callHMApiWithGateway`. I ran a [worker.py](https://linear.app/hivemapper/issue/EDGE-1311/investigation-of-app-proxy#comment-f30d0d66) script on my device to simulate an app. I modified the timestamp to distinguish this from the previous PUT to the backend.
6. I confirmed that the worker was being used by checking its logs, and checking odc-api logs for ("Submitting to gateway"). 
7. I confirmed with the repl that this new value was being published
8. I reverted all changes except `callHMApiWithGateway`

harroyoc (approved):

PR #608, opened 2026-02-03
send upload mode to BK

HariHive:
This PR exposes FrameKM upload mode to BK.


src/sqlite/config.ts
```diff
@@ -106,6 +103,11 @@ export const initializeUploadMode = async (): Promise<void> => {
       console.error('Error stopping wpa_supplicant:', error);
     }
   }
+
+  const success = await sendUploadModeToBK(uploadMode);
```
mazar-farran: Should this only run if beekeeper is enabled?

HariHive: great callout, added

mazar-farran:
I'm not sure about this approach, but I don't see a better way to propagate information to beekeeper with our current architecture.

I wonder if it would be better to have a generic `POST /configUpdated` which will cause BK to re-read all configs. But this is out of scope

mazar-farran (approved):
lgtm

HariHive:
> I'm not sure about this approach, but I don't see a better way to propagate information to beekeeper with our current architecture.
> 
> I wonder if it would be better to have a generic `POST /configUpdated` which will cause BK to re-read all configs. But this is out of scope

agree

PR #632, opened 2026-03-12
CAP-87: reject app proxy requests when no app connected recently

jimmdd:
## Summary
- Tracks when `/pending` was last polled by the app proxy
- `gatewayFetch` now immediately throws when in APP mode and no app has called `/pending` in the last 15 minutes, avoiding guaranteed-to-fail proxy requests
- Adds unit tests for `isAppConnectedRecently`, timestamp tracking, and the new gatewayFetch rejection behavior

## Test plan
- [x] Existing gateway tests still pass (150/150)
- [x] New tests verify: no-app-connected rejection, stale connection rejection, boundary conditions, timestamp storage
- [ ] Manual: verify that requests succeed normally when app is actively polling `/pending`
- [ ] Manual: verify that requests fail fast with clear error when no app is connected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mazar-farran:
This looks right but I'm still not sure I want this. It will make the code more stateful and difficult to reason about. Marking as a draft for now

PR #635, opened 2026-03-17
Restart map-ai when video plugin is newly enabled [CAP-71]

jimmdd:
When a video-needing plugin (beekeeper-plugin or video-capture-plugin) is newly enabled via the user config, restart map-ai so it picks up `videoEnabled = True` and initializes the video pipeline.

## Changes
- Added `previouslyEnabledPlugins` snapshot before plugin state updates in `pluginOnUserConfigUpdated`
- After all plugin state changes, compares before/after to detect if a video plugin was just enabled
- If so, restarts map-ai via `systemctl restart map-ai`
- Adds `DashcamRestartMapAiForVideoPlugin` instrumentation event

## Context
Without this, if the plugin registry call enables a video plugin after map-ai has already started, map-ai would continue running without video recording until its next restart (crash/reboot). This ensures the video pipeline is initialized immediately when needed.

Related: Hivemapper/hivemapper-hdc-os#700


evanmoss (approved):
cool, I think you might want this either run on `previouslyEnabledPlugins` or before the other plugins are disabled *iff* this is running after the map ai pipeline has already started

jimmdd (commented):
Good call — this is already handled: `previouslyEnabledPlugins` is captured at line 131 *before* any `setPluginState` calls. The `hadVideoPlugin` check at line 176 uses that pre-update snapshot, while `hasVideoPlugin` uses a fresh read after all updates. So the before/after comparison is accurate.

PR #639, opened 2026-03-27
feat: add video-events proxy route for plugins

jimmdd:
## Summary
Adds a proxy endpoint so plugins can report video upload events to the backend through the ODC API, which automatically attaches the device session cookie for user authentication.

### Changes
- **`src/routes/plugin.ts`** — New `POST /:pluginName/video-events` route that forwards the payload to `callHmApi("/plugins/${pluginName}/video-events")`

### How it works
1. Plugin uploads a video to S3
2. Plugin POSTs metadata to `http://127.0.0.1:5000/api/1/plugin/video-capture-plugin/video-events`
3. ODC API attaches the authenticated session cookie and forwards to the backend
4. Backend creates a `PluginVideoEvent` record linked to the user

### Related PRs
- **network**: https://github.com/Hivemapper/network/pull/5245
- **bee-video-capture-plugin**: https://github.com/Hivemapper/bee-video-capture-plugin/pull/1


hiveclawgit (commented):
I found 3 issue(s).

- [P2] Route parameter `/:pluginName` may greedily match other routes (src/routes/plugin.ts:181)
- [P3] No input validation on `pluginName` (src/routes/plugin.ts:182)
- [NIT] Dead check for `pluginName` (src/routes/plugin.ts:184)

hiveclawgit (commented):
**Re-review** (changes since last review):

I found 2 issue(s).

- [P2] `/secrets/:pluginName` lacks the same input validation (src/routes/plugin.ts:138)
- [NIT] `response.json()` can throw if upstream returns non-JSON (src/routes/plugin.ts:202)

hiveclawgit (commented):
**Re-review** (changes since last review):

I found 3 issue(s).

- [P2] Broad route pattern may swallow other plugin routes (src/routes/plugin.ts:183)
- [P2] Missing `return` after validation error in existing code (src/routes/plugin.ts:110)
- [NIT] Request body forwarded without validation (src/routes/plugin.ts:197)

hiveclawgit (commented):
**Re-review** (changes since last review):

✅ Looks good. I did not identify a discrete regression in this diff.

mazar-farran (approved):
lgtm, thanks for catching those missing returns

I wonder if there's a better way to do this that involves less boilerplate code, but I think it's important to have a restricted interface between plugins and backend

mazar-farran:
I think down the line we need to use something like gRPC to communicate with odc-api, but that will be a lot of work

PR #643, opened 2026-04-03
[CAP-105] Update embeddings endpoint to read from SQLite

hiveclawgit:
## Summary
- Replace file-based JSON reading (`/data/embeddings-cache/`) with SQLite queries for the `GET /embeddings` endpoint
- Add `Embeddings` TypeORM entity, migration (with timestamp index), and `src/sqlite/embeddings.ts` query module
- Update route tests to mock the SQLite layer instead of `fs` operations
- Add dedicated `__tests__/sqlite/embeddings.test.ts` with 7 test cases covering filtering, sorting, and JSON parsing
- Register new entity/migration in `src/sqlite/index.ts` and add sqlite test to CI workflow

## Context
Part of the Scene Embeddings project. CAP-101 (write side) double-writes to both SQLite and JSON files during the gradual rollout. This PR updates the read side to query SQLite instead of reading JSON files from disk.

Response format is unchanged:
```json
[{
  "filename": "...",
  "timestamp_ms": 1715027100000,
  "data": {
    "lat": 37.89,
    "lon": -122.12,
    "embedding": [0.014, 0.0, ...]
  }
}]
```

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `npx jest __tests__/routes/embeddings.test.ts __tests__/sqlite/embeddings.test.ts` — 18 tests pass
- [x] Full test suite — all 192 tests pass (6 pre-existing failures in unrelated mock/config files)
- [x] ESLint — no errors
- [x] Prettier — passes
- [ ] Verify on device that SQLite-backed endpoint returns same response format as file-based version

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mazar-farran:
Closing this because this was not good. Follow up #644

PR #651, opened 2026-04-13
CAP-94 update query-by-timestamp-ms endpoint

tylerlu-hivemapper:
https://linear.app/hivemapper/issue/CAP-94/design-and-implement-a-way-for-bee-plugins-to-use-embeddings

## Summary
- `GET /recordings/video/query-by-timestamp-ms` now returns a `videos` field with `{filepath, timestamp_ms}` objects alongside the existing `files` field (bare path strings)
- Existing `files` field unchanged — no breaking change for current consumers

## Test plan
- [x] `npx tsc --noEmit` passes


mazar-farran:
This is used here https://github.com/Hivemapper/beekeeper-plugin-service/blob/main/src/util/latestMedia.ts#L71, and possibly by other plugins

mazar-farran (approved):

